PR #40, opened 2016-07-08
Fix Dockerfile to work again

gregkare:
This now works on my machine (TM). It took me way too long because I don't know how to Docker


Dockerfile
```diff
@@ -1,14 +1,23 @@
-FROM nodesource/node:trusty
+FROM nodesource/trusty:4.4.7
```
bkero: Do we want to pin to a particular nodesource version? Would 'latest' work here, or 5?

https://hub.docker.com/r/nodesource/trusty/

Dockerfile
```diff
@@ -1,14 +1,23 @@
...
 MAINTAINER Ben Kero <ben.kero@gmail.com>
 
-ADD package.json /tmp/package.json
 RUN apt-get update && apt-get install -y git-core
-RUN cd /tmp && npm install
-RUN mkdir -p /hyperchannel && cp -a /tmp/node_modules /hyperchannel/
+
+# The Docker image now sets NODE_ENV="production". All our dependencies are
+# devDependencies
+ENV NODE_ENV development
```
bkero: I would suggest putting this near the command that this environment variable affects

skddc:
The error message for the failed build is rather useless. But I dont' think changing the Dockerfile can break the Ember test suite.

gregkare:
I didn't see the failed build. I restarted it, it's doing the same thing (nothing happens for 10 minutes, Travis cancels it)

gregkare:
Now that the Sockethub client is downloaded from Sockethub [the hardcoded localhost Sockethub](https://github.com/67P/hyperchannel/blob/6bbe907bcd3413fdcca45c33515d8ae29273e6ab/app/index.html#L21-L23) cannot work inside of Docker (since sockethub is running in another container). I'm going to look at how to make this configurable (per environment maybe?)

skddc:
> Now that the Sockethub client is downloaded from Sockethub the hardcoded localhost Sockethub cannot work inside of Docker (since sockethub is running in another container)

I thought that's what container linking is for. Pretty sure this is not just possible, but a usual scenario with Docker Compose.

gregkare:
@skddc Loading it through localhost doesn't work if you're using Docker Machine (you need to manually redirect ports so it's really not convenient). Loading those JS files from localhost should however work on Linux, can you run `docker-compose up` in the orchestration repo?

I need to set up a Linux VM for "native" Docker support but it's using so much disk space I need to clear some space on my laptop first

skddc:
Any news on this one? Should I still test this somehow? (Have to fix my Docker daemon since moving to a new machine broke it somehow).

gregkare:
We should probably wait until we're not on LTE anymore to test this again

skddc:
I bought enough bandwidth. But that wasn't the question anyway.

gregkare:
I'm going to look at how this works with the current master

skddc:
Whoever just wants to run this locally with a remote Sockethub can use `EMBER_ENV=production npm start` by the way. I'm not sure we need Hyperchannel running within Docker for development, and we deploy it to 5apps Deploy for production.

skddc:
Closing this PR for now, leaeving the branch intact.

PR #43, opened 2016-10-14
Update to ember-cli and Ember 2.8.0

galfert:
Also loads the Sockethub client directly from the Sockethub server instead of using `ember-cli-sockethub`.


app/index.html
```diff
@@ -7,23 +7,24 @@
...
+    <link rel="stylesheet" href="{{rootURL}}assets/hyperchannel.css">
 
-    {{content-for 'head-footer'}}
+    {{content-for "head-footer"}}
   </head>
   <body>
-    {{content-for 'body'}}
+    {{content-for "body"}}
 
     <!-- TODO these sockethub dependencies should be loaded via ember-cli-sockethub instead -->
```
skddc: "should be loaded from configured Sockethub server" perhaps?

skddc:
`ember test` (and `npm test`) is currently failing for me with the following message:

> Using a version of ember-cli-shims prior to 0.1.0 will cause errors while loading Ember Data 2.3.0-beta.3+. Please update ember-cli-shims from 0.0.3 to 0.1.0.

PR #47, opened 2016-10-14
Scroll to bottom of chat window when focussing channel

skddc:
I found out how to scroll to the bottom of our chat messages container in general. I also added that to the route hook, which is focussing the message input when switching to a channel.

However, I think the `afterRender` hook probably belongs in a channel component, same as the scroll function call when adding new messages. What do you think, @galfert?


galfert:
Yes, a component would be a better place for this.

galfert:
Works and looks all good.

The scroll animation is indeed a bit much. Especially when switching channels. But just setting the animation time to 0 doesn't improve it for me.

skddc:
Let's use it for a bit and see if we want to change it.

PR #48, opened 2016-10-14
[WIP] Add yarn.lock and document Yarn in readme

galfert:
Currently when running `yarn install` (or when adding a new module with `yarn add <package`) it will delete all contents in the `bower_components` directory. There is (at least) one open bug issue about it at https://github.com/yarnpkg/yarn/issues/616.

Running `bower install` will install the missing packages again.

Also running the commands defined in the `package.json` like `build`, `start`, `test`, ... via yarn throws an error for me:

``` bash
$ yarn test
yarn test v0.15.1
$ "ember test"
sh: ember test: command not found
error Command failed with exit code 127.
info Visit http://yarnpkg.com/en/docs/cli/test for documentation about this command.
```


galfert:
At least the problem with running commands (e.g. `yarn start`) has been fixed in the latest version (i.e. 0.16.0).

skddc:
Is this something we want to keep open, i.e. are we considering to switch to Yarn anytime soon? I'm a bit uncomfortable with what I've seen so far tbh.

galfert:
> I'm a bit uncomfortable with what I've seen so far tbh.

I'm feeling the exact same way. I don't think this is worth keeping open at this time.

PR #49, opened 2016-10-28
Refactoring

galfert:
Some refactoring action to clean up a little bit and keep on par with Ember best practices:
- Move actions from `base_channel`-route to `base_channel`-controller. As there were too many controller concerns in there, they were better suited in the controller. Also it's needed for closure actions (see below) to work.
- Use [closure actions](https://spin.atomicobject.com/2016/06/25/emberjs-closure-actions/) for the `channel-container` component. One of the benefits of this is that you can directly call the action method and use return values or Promises this way.
- Remove `smt` initializer and instead inject the service explicitly where needed


skddc:
Cool stuff!

I tested it real quick and found that joining new channels via `/join` and changing topic via `/topic` doesn't work anymore. Curiously, leaving channels via `/part` still works, so I'm not sure what's broken there.

galfert:
The `/join` command had a bug indeed. That should be fixed now. Thanks for testing!

I couldn't see any problem with the `/topic` command though. Did you maybe try to set the topic for a channel that you didn't have operator status for? In that case there is an error in the sockethub console. But it seems that the error isn't reported back to the app, that's why there is no feedback in Hyperchannel (yet).

Setting the topic for a channel you are an operator in should work as expected.

skddc:
> The /join command had a bug indeed. That should be fixed now. Thanks for testing!

Yup, works now. Great.

> Setting the topic for a channel you are an operator in should work as expected.

That was indeed the problem. Thanks.

Issue #45, opened 2016-10-14
Order room list alphabetically


skddc:
When joining a new room, it's appended to the end of the list currently.

galfert:
The PR implementing this had been merged quite a while ago.

PR #54, opened 2016-12-02
Sort the channels in the sidebar

gregkare:
Refs #45 

@galfert Would you add a test for something like this?

### DIFF
--- a/app/models/space.js
+++ b/app/models/space.js
@@ -25,6 +25,9 @@ export default Ember.Object.extend({
 
   sockethubPersonId: function() {
     return `irc://${this.get('ircServer.nickname')}@${this.get('ircServer.hostname')}`;
-  }.property('ircServer.hostname', 'ircServer.nickname')
+  }.property('ircServer.hostname', 'ircServer.nickname'),
+
+  channelSorting: ['name'],
+  sortedChannels: Ember.computed.sort('channels', 'channelSorting')
 
 });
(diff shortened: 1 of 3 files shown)


app/models/space.js
```diff
@@ -25,6 +25,9 @@ export default Ember.Object.extend({
 
   sockethubPersonId: function() {
     return `irc://${this.get('ircServer.nickname')}@${this.get('ircServer.hostname')}`;
-  }.property('ircServer.hostname', 'ircServer.nickname')
+  }.property('ircServer.hostname', 'ircServer.nickname'),
+
+  channelSorting: ['name'],
+  sortedChannels: Ember.computed.sort('channels', 'channelSorting')
```
skddc: Any reason this has to be a seperate variable here? It's actual making the same line longer than without one.

skddc:
thumbs-up-cool2

app/models/space.js
gregkare: Unless I'm missing something the second argument has to be an attribute as a string or a function (http://emberjs.com/api/classes/Ember.computed.html#method_sort)

skddc: Ah, didn't know that. Thanks.

galfert:
Nice one.

I added a test for the channel sorting.

gregkare:
Thanks for the test!

PR #55, opened 2016-12-09
Render links in topics

### LINKED ISSUE #44: Render links in topics
Similar to #18

gregkare:
Fixes #44

### DIFF
--- a/app/models/channel.js
+++ b/app/models/channel.js
@@ -23,8 +23,8 @@ export default Ember.Object.extend({
       });
       return new Ember.String.htmlSafe(topic);
     } else {
-      return "";
+      return '';
     }
-  }.property('message'),
+  }.property('topic')
 
 });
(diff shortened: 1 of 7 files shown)


app/models/channel.js
```diff
@@ -15,4 +15,16 @@ export default Ember.Object.extend({
     return this.get('name').replace(/#/g,'');
   }.property('name'),
 
+  formattedTopic: function() {
+    if (this.get('topic') !== null) {
+      let topic = linkifyStr(this.get('topic'), {
+        defaultProtocol: 'https',
+        linkAttributes: { rel: 'nofollow' }
+      });
+      return new Ember.String.htmlSafe(topic);
```
galfert: Not sure this is a good idea. Marking the whole `topic` as `htmlSafe` could allow XSS.

galfert:
Tried this out and works as advertised.

I also fixed the dependent key for the property and added some tests.

So this should be all good to merge now.

gregkare:
Thanks

Issue #46, opened 2016-10-14
Indicate channel activity in channel list


skddc:
The channel list should indicate activity for non-focussed channels.
- [x] Add class to element on activity
- [x] Remove class when focussing channel
- [x] Style unread-messages channel list item

skddc:
I wonder if we should start with a read status right away, because that's what we're going to need anyway. I mean instead of just using "focussing the channel".

galfert:
The number of unread messages next to the channel would be nice as well.

skddc:
I'm not sure about that. Do you really have that in *any* team chat app you use or know? I think it migh be too much information (i.e. cluttering the UI with something that's not useful and takes away brain cycles from the important data). Also, I would assume actual implementation problems for getting that number right with e.g. IRC.

skddc:
... in any case, if we add read status to the model, it'd be very easy to experiment with that, or even make it configurable.

PR #56, opened 2017-01-12
Unread status for channels

skddc:
This adds a visibility status to channels, and marks them as unread when new messages arrive, while they're not visible.

Connected to #46

### DIFF
--- a/app/models/channel.js
+++ b/app/models/channel.js
@@ -8,6 +8,9 @@ export default Ember.Object.extend({
   connected: false,
   sockethubChannelId: null,
   topic: null,
+  unreadMessages: false,
+  unreadMentions: false,
+  visible: false, // Current/active channel
 
   slug: function() {
     // This could be based on server type in the future. E.g. IRC would be
@@ -25,6 +28,13 @@ export default Ember.Object.extend({
     } else {
       return '';
     }
-  }.property('topic')
+  }.property('topic'),
+
+  unreadMessagesClass: function() {
+    if (this.get('visible') || !this.get('unreadMessages')) {
+      return null;
+    }
+    return this.get('unreadMentions') ? 'unread-mentions' : 'unread-messages';
+  }.property('visible', 'unreadMessages', 'unreadMentions')
 
 });
--- a/app/components/channel-container/component.js
+++ b/app/components/channel-container/component.js
@@ -1,5 +1,15 @@
 import Ember from 'ember';
 
+function scrollToBottom() {
+  Ember.$('#channel-content').animate({
+    scrollTop: Ember.$('#channel-content ul').height()
+  }, '500');
+}
+
+function focusMessageInput() {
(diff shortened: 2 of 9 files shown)


skddc:
Sorry for the typo in the branch name. ;)

app/models/channel.js
```diff
@@ -8,6 +8,9 @@ export default Ember.Object.extend({
   connected: false,
   sockethubChannelId: null,
   topic: null,
+  unreadMessages: false,
+  unreadMentions: false,
+  visible: false, // Current/active channel
```
galfert: Not sure `visible` is a good choice. Shouldn't it rather be `active`?

`visible` makes me think you could also hide channels somehow.

galfert:
Works good.

I added the missing code for marking the channel when receiving unread mentions.

skddc:
Nice!

app/models/channel.js
skddc: I thought about it and at the time decided that "visible" was more apt. Thinking about it again, it's all the same to me. I just didn't like how "active" made it sound like other channels were somehow inactive, when this very feature is about indicating channel activity in channels that are not currently open/visible.

silverbucket:
What about notify / notification?

On Sat, Jan 14, 2017, 13:33 Sebastian Kippe <notifications@github.com>
wrote:

*@skddc* commented on this pull request.
------------------------------

In app/models/channel.js <https://github.com/67P/hyperchannel/pull/56>:

> @@ -8,6 +8,9 @@ export default Ember.Object.extend({
   connected: false,
   sockethubChannelId: null,
   topic: null,
+  unreadMessages: false,
+  unreadMentions: false,
+  visible: false, // Current/active channel

I thought about it and at the time decided that "visible" was more apt.
Thinking about it again, it's all the same to me. I just didn't like how
"active" made it sound like other channels were somehow inactive, when this
very feature is about indicating channel activity in channels that are not
currently open/visible.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/67P/hyperchannel/pull/56>, or mute the thread

skddc:
`notify` could be a computed property, determining if notifications should be shown based on both visibility of the channel as well as per-channel notification settings. `visible` is only used to determine if the channel is open/visible/active/current or not.

(By the way, your whole email was copied to the GitHub comment. You might want to check your settings for reply format.)

silverbucket:
Oh yeah I forgot github reply from email does that.

Maybe `updated` flag? to indicate that there have been updates to the
channel since you last looked?

skddc:
This is all implemented already, and it's using `unreadMessages` and `unreadMentions` as properties on the channel model for that.

skddc:
Added some tests for the unread class function, but we'll have to look into integration tests for what we have as well as future features asap.

PR #57, opened 2017-01-16
Render links in messages

### LINKED ISSUE #18: Render links in text messages

galfert:
Closes #18

### DIFF
--- a/tests/unit/components/message-chat-test.js
+++ b/tests/unit/components/message-chat-test.js
@@ -12,7 +12,7 @@ test('#formattedContent turns full URLs into links', function(assert) {
   assert.equal(component.get('formattedContent').toString(), 'visit <a href="https://kosmos.org" class="linkified" target="_blank" rel="nofollow">https://kosmos.org</a> for more info');
 });
 
-test('#formattedContent ignores URLs without protocol', function(assert) {
+test('#formattedContent does not turn domain names into links', function(assert) {
   let component = this.subject({
     message: { content: 'visit kosmos.org for more info' }
   });
--- a/tests/unit/components/message-chat-test.js
+++ b/tests/unit/components/message-chat-test.js
@@ -1,0 +1,22 @@
+import { moduleForComponent, test } from 'ember-qunit';
+
+moduleForComponent('message-chat', 'Unit | Component | message-chat', {
+  unit: true
+});
+
+test('#formattedContent turns URLs into links', function(assert) {
+  let component = this.subject({
+    message: { content: 'visit kosmos.org for more info' }
+  });
+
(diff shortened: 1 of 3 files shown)


tests/unit/components/message-chat-test.js
```diff
@@ -0,0 +1,22 @@
+import { moduleForComponent, test } from 'ember-qunit';
+
+moduleForComponent('message-chat', 'Unit | Component | message-chat', {
+  unit: true
+});
+
+test('#formattedContent turns URLs into links', function(assert) {
+  let component = this.subject({
+    message: { content: 'visit kosmos.org for more info' }
```
skddc: This is not actually a URL being turned into a link, but something that resembles an FQDN. I think we should only link actual URLs, because we can't know if there's anything behind that domain name, much less if it's via HTTPS.

tests/unit/components/message-chat-test.js
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  assert.equal(component.get('formattedContent').toString(), 'visit <a href="https://kosmos.org" class="linkified" target="_blank" rel="nofollow">kosmos.org</a> for more info');
+});
+
+test('#formattedContent escapes HTML', function(assert) {
```
skddc: 👍

tests/unit/components/message-chat-test.js
galfert: Ok, I will change that.
Btw, I think waves is doing the same thing.

galfert:
Changed it so that only full URLs that include a protocol will be rendered as links.

tests/unit/components/message-chat-test.js
```diff
@@ -4,12 +4,20 @@ moduleForComponent('message-chat', 'Unit | Component | message-chat', {
...
-test('#formattedContent turns URLs into links', function(assert) {
+test('#formattedContent turns full URLs into links', function(assert) {
+  let component = this.subject({
+    message: { content: 'visit https://kosmos.org for more info' }
+  });
+
+  assert.equal(component.get('formattedContent').toString(), 'visit <a href="https://kosmos.org" class="linkified" target="_blank" rel="nofollow">https://kosmos.org</a> for more info');
+});
+
+test('#formattedContent ignores URLs without protocol', function(assert) {
```
skddc: I realize it's a bit picky, but technically this is just a domain name or FQDN, not a "URL without protocol". ;)

https://en.wikipedia.org/wiki/Uniform_Resource_Locator#Syntax

galfert: It might contain a path as well, then it wouldn't be just a domain name.

skddc: That still wouldn't make it a URL though.

skddc:
Sorry again. ;)

tests/unit/components/message-chat-test.js
galfert: Changed it.

PR #59, opened 2017-01-21
Deployment

skddc:
I set up deployment to 5apps Deploy via scripts, using sockethub.kosmos.org in production. All working. Just waiting for DNS to propagate for https://hyperchannel.kosmos.org/


silverbucket:
Woo-hoo! \o/

PR #61, opened 2017-01-22
Basic mobile support

skddc:
This hides the sidebars on small screens and opens them via swipe left/right or click on a channel's title bar. It also hides the channel titles until we've implemented a better way to expand long entries.

I added an addon that implements a helper for sending actions to routes directly. I also implemented a generic application route action for sidebar menu handling. See code for usage/details.

(This branch is currently deployed to https://hyperchannel.kosmos.org)


skddc:
Had a pretty good hacking session. It looks and feels quite nice now, except for the swiping not working in Chrome, when a channel is open (and in reverse the title bar tapping not working in Webkit then).

Now the biggest issue is the missing history.

galfert:
Very nice.

I added the closing of the menu when tapping outside of it.

I'd say this is good to merge now.

PR #62, opened 2017-01-23
Load today's channel logs

galfert:
Refs #4 

This loads today's logs when joining a channel from the public Kosmos storage when available.

Also did some refactorings along the way.


skddc:
hyperdunk

PR #64, opened 2017-01-23
Render images inline

### LINKED ISSUE #23: Render images inline
Based on #22.

skddc:
(currently deployed to prod)

closes #23
connected to #23

### DIFF
--- a/app/components/message-chat/component.js
+++ b/app/components/message-chat/component.js
@@ -7,16 +7,28 @@ export default Ember.Component.extend({
 
   formattedContent: function() {
     const content = this.get('message.content');
+    let out;
 
-    let out = linkifyStr(content, {
-      linkAttributes: { rel: 'nofollow' },
-      validate: {
-        url: function (value) {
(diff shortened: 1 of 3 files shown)


app/components/message-chat/component.js
```diff
@@ -7,16 +7,28 @@ export default Ember.Component.extend({
...
+    let out;
 
-    let out = linkifyStr(content, {
-      linkAttributes: { rel: 'nofollow' },
-      validate: {
-        url: function (value) {
-          return /^(http)s?:\/\//.test(value);
+
+    // Images
+    if (content.match(/^(http)s?:\/\/.*(gif|png|jpg|jpeg)$/)) {
```
galfert: Shouldn't the regex be case insensitive so extensions like `.JPG` would work as well?

skddc: Yes, good idea.

skddc: Done.

PR #65, opened 2017-01-23
Feat(message-chat): Add date

fsmanuel:
@skddc review?

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "ember-moment": "7.3.0",
     "ember-resolver": "^2.0.3",
     "ember-route-action-helper": "2.0.2",
-    "ember-welcome-page": "^1.0.1",
     "loader.js": "^4.0.1"
   }
 }
--- a/README.md
+++ b/README.md
@@ -26,9 +26,10 @@ You will need the following things properly installed on your computer.
 
 ## Installation
 
-* `git clone <repository-url>` this repository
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -52,13 +52,11 @@
     "ember-cli-uglify": "^1.2.0",
     "ember-export-application-global": "^1.0.5",
     "ember-gestures": "0.4.2",
-    "ember-getowner-polyfill": "1.1.1",
```
galfert: I think this library is currently a dependency of `ember-gestures`. Are you sure this can be removed?

fsmanuel: Good catch. I got the deprecation that it is in Ember since 2.3. I checked `ember-gestures` and yeah they just have it in the devDependencies. Which is wrong cause the idea of the polyfill is to polyfill for addons if they don't know what version of ember the consuming app uses. I'll readd it just to make sure nothing breaks.

fsmanuel:
@galfert I squashed the commits but I don't know why @skddc changes are in the diff...

skddc:
> @galfert I squashed the commits but I don't know why @skddc changes are in the diff...

That's because he merged https://github.com/67P/hyperchannel/pull/64 which changed some of the same code.

skddc:
Merged locally in fd94d30ace6831228e124dbe0373a1b255791489. Thanks!

PR #66, opened 2017-01-23
Feat(channel): Default channel and persist currentSpace and currentChannel

fsmanuel:
@galfert I added freenode#kosmos as the default channel. The PR persists the `currentSpace` and `currentChannel` into localStorage and redirects to it.

/cc @skddc


fsmanuel:
@galfert We can also delete `index.hbs` because it's no longer in use...

skddc:
Rebased and merged manually.

PR #68, opened 2017-01-23
Feat(logger): Add a logger service

### LINKED ISSUE #67: Logger service
We should add a logger service to reduce the console noise.

fsmanuel:
I would propose something like that to close #67
Any feedback is welcome!

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,11 +1,26 @@
 import Ember from 'ember';
 
-export default Ember.Route.extend({
+const {
+  Route,
+  get,
+  inject: {
+    service
+  }
+} = Ember;
 
-  smt: Ember.inject.service(),
+export default Route.extend({
+  logger: service(),
+  smt: service(),
 
-  model: function() {
-    this.get('smt').loadFixtures();
+  beforeModel() {
+    // See a list of allowed types in logger.js
+    // Add or remove all your log types here:
+    // get(this, 'logger').addToLogs('message');
+    // get(this, 'logger').removeFromLogs('join');
+  },
+
+  model() {
+    get(this, 'smt').loadFixtures();
   },
 
   actions: {
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -12,6 +12,7 @@ export default Route.extend({
   smt: service(),
 
   beforeModel() {
+    this._super(...arguments);
     // See a list of allowed types in logger.js
     // Add or remove all your log types here:
     // this.get('logger').addToLogs('message');
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -15,8 +15,10 @@ export default Route.extend({
     this._super(...arguments);
     // See a list of allowed types in logger.js
     // Add or remove all your log types here:
-    // this.get('logger').addToLogs('message');
-    // this.get('logger').removeFromLogs('join');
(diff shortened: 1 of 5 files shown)


app/services/logger.js
```diff
@@ -0,0 +1,48 @@
...
+    'send'
+  ],
+
+  log(type) {
+    if (!get(this, 'allowedTypes').includes(type)) {
+      throw new Error(`You specified a unknown type: "${type}".`);
+    }
+
+    if (get(this, 'showLogsFor').includes(type)) {
+      console.log.apply(null, arguments);
```
skddc: I think this should use `Ember.Logger` instead.

app/services/smt.js
```diff
@@ -49,7 +50,7 @@ export default Ember.Service.extend({
       }
     };
 
-    Ember.Logger.debug('connecting to irc', credentials);
+    this.get('logger').log('connection', 'connecting to irc', credentials);
```
skddc: Maybe we can mix this in somehow so it's just `this.log()`?

app/services/logger.js
```diff
@@ -0,0 +1,48 @@
...
+    get(this, 'showLogsFor').push(type);
+  },
+
+  removeFromLogs(type) {
+    let types = get(this, 'showLogsFor');
+    let index = types.indexOf(type);
+
+    if (index > -1) {
+        types.splice(index, 1);
+    }
```
galfert: Couldn't you just use Ember's `removeObject` here?

fsmanuel:
@skddc @galfert updated

app/routes/application.js
```diff
@@ -8,8 +8,16 @@ const {
 } = Ember;
 
 export default Route.extend({
+  logger: service(),
   smt: service(),
 
+  beforeModel() {
```
galfert: You should at least call `this._super(...arguments)` here.

fsmanuel: Good catch!

fsmanuel: updated

app/routes/application.js
```diff
@@ -8,8 +8,17 @@ const {
...
+    this._super(...arguments);
+    // See a list of allowed types in logger.js
+    // Add or remove all your log types here:
+    // this.get('logger').addToLogs('message');
+    // this.get('logger').removeFromLogs('join');
```
skddc: What do you think about adding a line that would remove all logs at once, when uncommenting it?

fsmanuel: I changed the public api to `.add(type)` `.remove(type)` and added `.disable()` and `.enable()` + tests

fsmanuel:
@skddc @galfert updated

skddc:
Very cool!

I added the remaining log statements (which I stilll saw when logs weren't disabled) and formatted the type in the output with brackets.

I think it's good to merge now.

fsmanuel:
👍

galfert:
Looking good 👍 

Btw, you can access the logger during runtime from the console via

`window.Hyperchannel.__container__.lookup('service:logger')`

skddc:
> Btw, you can access the logger during runtime from the console via
> 
> window.Hyperchannel.__container__.lookup('service:logger')

Nice, thanks!

PR #70, opened 2017-01-24
Add remoteStorage 1.0.0-beta from npm module

skddc:
This adds the current beta version directly from the RS repo's noglobal branch and imports it via ember-browserify as ES6 module.

It also adds the RS instance to `window` so it's available for debugging and playing around in the browser console.


skddc:
Looks like we have to teach the test suite how to deal with this way of importing a module.

Issue #63, opened 2017-01-23
Menu swiping not working in Chrome


skddc:
Swiping on the channel container area to open/close the menu works in Mobile Safari and Firefox, but not in Chrome.

skddc:
Chrome desktop with mobile emulation actually does work for me, so it looks like a bug in my Chrome for Android.

skddc:
Note: same issue in all of stable, beta and canary on Android.

skddc:
I tried to add the swipe handlers to the `channel-container` component, but even there no swipes are even registered, while the `app-container` swipes work just fine as long as the `channel-container` isn't open. I'm kind of at the end of my abilities to debug this. @galfert Any ideas maybe?

skddc:
Here's the best part, uncovered today:

In Safari, tapping the channel header doesn't work for opening/closing the menu, so it's exactly the other way around. :)

galfert:
I found that at least in mobile Chrome the`overflow: auto` on the `.main` class (https://github.com/67P/hyperchannel/blob/9df59f07ff8e099371f5c3fc9d7d40b8e37b24a9/app/styles/_layout.scss#L13) is what prevents the swipe events.

Haven't checked Safari yet.

galfert:
I was able to at least fix the Safari problem.

There seems to be a bug that requires an empty click handler on the `header` element in order for the `menu` action to get triggered  `¯\_(ツ)_/¯`

galfert:
Here's the corresponding bug report in Hammer.js: https://github.com/hammerjs/hammer.js/issues/1065

skddc:
> I was able to at least fix the Safari problem.

Yay!

> Here's the corresponding bug report in Hammer.js: hammerjs/hammer.js#1065

Cool. Looks like they kinda have to fix it asap, because everyone's running into it.

galfert:
One of the contributors just commented on the issue:

> I think this is due to chrome making touchEvents passive by default. Ill try to find time to lookinto this and fix this week

\o/

For the meantime I created a temporary fix at #71.

PR #71, opened 2017-01-26
Fix swiping in Chrome

galfert:
Connected to #63

I created an intermittent patched version of hammer.js that fixed our swiping problem in mobile Chrome. I got that fix from the original issue discussion. Apparently it breaks pan support in desktop Chrome, but since we are not using that I think that's ok for now.

I'll keep an eye on the original issue to see when they found a proper fix for it.

While at it I also enabled momentum scrolling for mobile Safari. Feels much nicer.

### DIFF
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -11,6 +11,7 @@
   .main {
     @include flex(1);
     overflow: auto;
+    -webkit-overflow-scrolling: touch;
   }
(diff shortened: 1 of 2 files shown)


skddc:
Hmm, this doesn't work in my Chrome with mobile emulation active. Up/down swiping and tapping works in there. How did you test this?

PR #72, opened 2017-01-26
Don't bounce when scrolling to top or bottom on mobile Safari

galfert:
This keeps the top menu at it's place when scrolling all the way up to the top.


skddc:
Nice! That looked super annoying when I watched someone use it. ;)

PR #73, opened 2017-01-27
Use remoteStorage for persistence

skddc:
This adds the new beta branch of remoteStorage.js as well as a very simple Kosmos module for managing spaces. It removes the existing fixtures and replaces them with a default space stored in RS.

This is all working nicely, but we need to figure out how to use rs.js in the tests.


app/services/remotestorage.js
```diff
@@ -0,0 +1,57 @@
...
+  rsInstance: null,
+
+  rs: function() {
+    if (this.get('rsInstance')) { return this.get('rsInstance'); }
+
+    let rs =  new RemoteStorage(/* {logging: true} */);
+    rs.access.claim('kosmos', 'rw');
+    rs.caching.enable('/kosmos/');
+    return rs;
+  }.property('rs'),
```
galfert: I think this is supposed to depend on `rsInstance`, not `rs`.

Also, after creating a new instance, it should be saved as `rsInstance`.

app/models/space.js
```diff
@@ -2,38 +2,51 @@ import Ember from 'ember';
...
     password: null,
     nickname: null,
     nickServ: {
       password: null
     }
   },
-  channels  : null,
-  users     : null,
+  channels   : null, // Channel instances
+  channelList: [],   // Bookmarked channel names
```
galfert: You should never set an `Ember.Object`'s property to an array like this, because then the same array will be shared by all instances of the object.

You should rather set the initial empty array in the `init()` method instead.

See also https://guides.emberjs.com/v2.6.0/object-model/classes-and-instances/#toc_initializing-instances

app/controllers/space/base_channel.js
```diff
@@ -43,21 +44,26 @@ export default Ember.Controller.extend({
...
-        console.log('Unknown command', commandText);
+        Ember.Logger.warn('[channel]', 'Unknown command', commandText);
       }
 
       this.set('newMessage', null);
     },
 
     joinCommand: function(args) {
+      let space = this.get('space.model');
       let channel = this.get('smt').createChannel(this.get('space.model'), args[0]);
```
galfert: You could use the new `space` variable from above here as well.

app/app.js
```diff
@@ -7,7 +7,11 @@ let App;
 
 Ember.MODEL_FACTORY_INJECTIONS = true;
 
+Ember.deprecate = function(){};
+Ember.warn = function(){};
```
galfert: Don't think it's a good idea to silence these.

app/app.js
```diff
@@ -7,7 +7,11 @@ let App;
...
+
 App = Ember.Application.extend({
+  LOG_TRANSITIONS: true,
```
galfert: Did you intend to leave this in?

app/controllers/application.js
```diff
@@ -12,10 +12,15 @@ const {
 
 export default Controller.extend({
   smt: service(),
+  storage: service('remotestorage'),
   spaces: alias('smt.spaces'),
 
+  // Make the RS instance available in the browser console for development
+  setRemoteStorageGlobal: function() {
+    window.remoteStorage = this.get('storage.rs');
+  }.on('init'),
```
galfert: You might want to do this from `app/initializers/remotestorage.js` instead.

skddc: I couldn't use the service from there.

app/services/remotestorage.js
skddc: Right, I "refactored" this away in a rush last night. :)

app/controllers/application.js
galfert: Oh, right. For that it actually needs to be an instance initializer in `app/instance-initializers/` instead. The you can do something like:

```
initialize(instance) {
  let storage = instance.lookup('service:remotestorage');
  ...
```

skddc:
Turns out PhantomJS stable doesn't support ES6 (and thus Promises), while the rs.js branch doesn't require `Promise` from the bluebird package where it's using it. So it works in browsers and node, but not PhantomJS.

They have a 2.5 beta with ES6 support, but we can't use that everywhere yet.

skddc:
I added an addon, which loads Babel polyfills in the test environment. Tests are green now. Let's get this merged asap, because it's the basis for many other features.

skddc:
There was another issue, which was `xmlhttprequest` missing for the browserification of remoteStorage.js (in order to be used via ES6 import). I think it should be part of `dependencies` instead of `devDependencies`. Will ask in the rs.js PR.

All green now. 🎉

galfert:
Not sure what's going on exactly yet, but I don't get joined into any channels anymore using this branch.

There is no errors in the console, channels are shown as connected and logs are loaded as well. But I don't see any users and don't get new messages, nor do my messages actually arrive in the channel. Using a normal IRC client I also don't see any notification about my hyperchannel user joining.

Pretty sure it worked before the recent changes. And current master also still works, so doesn't appear to be Sockethub's fault.

skddc:
What's being logged in your console? Unfortunately, I run into issues with Sockethub when reloading the app, switching branches, etc. rather frequently.

skddc:
What's being logged in your console? Unfortunately, I run into issues with Sockethub when reloading the app, switching branches, etc. rather frequently.

Edit: I just checked again and it works fine over here. Pretty sure the changes don't affect IRC connections.

galfert:
Weird, never had those issues before. And I restarted Sockethub between switching branches.

One thing I can see in the logs is that it tries to connect to the server twice and and also joining all channels twice:

log-1

Asking for attendance list even happens 4 times per channel (see e.g. for the #hackerbeach channel, 2 logs at the top, 2 more at the bottom):

log-2

log-3

skddc:
Could be that the routine is failing when adding the default space. Is it the same after you added the default space?

galfert:
It's the same after a reload when the space config is coming from RS.

skddc:
I don't see how it can happen with either scenario from the code. Did you try other browsers/windows yet? Maybe something is cached in Sockethub's Redis across restarts?

galfert:
Tried different browser and also purged all sockethub keys from Redis. Still the same.

galfert:
Found the culprit to the be `afterModel` hook in the `application` route. It caused the `model` hook to be called a second time, which caused the double creation of the same space and channels and also set up two Sockethub listeners. That's why I saw the attendance list message being sent 4 times per channel (both of the listeners sent out the message for both of the channels).

And creating the same space twice apparently also confused Sockethub, so I didn't get connected at all.

I moved the logic from the `application` route's `afterModel` to the `application.index` route's `redirect` hook. There it makes more sense anyway, because we don't need to redirect to the last used channel when coming in to a channel route right away.

silverbucket:
Cool, could you post a sockethub issue with the steps to confuse it so I can test for that and perhaps provide better feedback?

skddc:
Great!

Good to merge now?

galfert:
Yep, good to merge. Great job, @skddc.

sanders_thumbs_up

galfert:
@silverbucket I will produce a test case to produce the problems in Sockethub for you.

skddc:
thanks-ht

PR #78, opened 2017-01-28
Join channels via sidebar link and prompt

skddc:
Adds a "+" link as last item of the channel nav and will transition/join a channel name given via prompt.

### DIFF
--- a/app/components/channel-nav/component.js
+++ b/app/components/channel-nav/component.js
@@ -2,14 +2,6 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
 
-  tagName: 'ul',
-
-  actions: {
-
-    joinChannel(space) {
-      this.sendAction('joinChannel', space.get('id'));
-    }
-
-  }
+  tagName: 'ul'
 
 });
--- a/app/components/channel-nav/component.js
+++ b/app/components/channel-nav/component.js
@@ -2,7 +2,14 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
 
-  tagName: 'ul'
+  tagName: 'ul',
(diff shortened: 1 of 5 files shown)


app/components/channel-nav/template.hbs
```diff
@@ -2,15 +2,18 @@
...
+        <li class="{{if channel.connected "connected" "disconnected"}} {{channel.unreadMessagesClass}}">
+          {{#if channel.isUserChannel}}
+            {{#link-to "space.userChannel" space channel}}{{channel.name}}{{/link-to}}
+          {{else}}
+            {{#link-to "space.channel" space channel}}{{channel.name}}{{/link-to}}
+          {{/if}}
+        </li>
+      {{/each}}
+      <li>
+        <a {{action "joinChannel" space}} class="join-channel">+</a>
```
galfert: If you do 

```
<a {{action (action joinChannel space.id)}} class="join-channel">+</a>
```

instead, you can get rid of the whole action definition in the component itself.

Action closures FTW :)

app/components/channel-nav/component.js
```diff
@@ -2,7 +2,14 @@ import Ember from 'ember';
...
-  tagName: 'ul'
+  tagName: 'ul',
 
+  actions: {
+
+    joinChannel(space) {
+      this.sendAction('joinChannel', space.get('id'));
+    }
+
+  }
```
galfert: No need for this, see comment for the template.

app/components/channel-nav/template.hbs
skddc: Wow, nice! :)

PR #79, opened 2017-02-01
Add ServiceWorker for offline assets

skddc:
This adds `broccoli-serviceworker`, which automatically generates a `ServiceWorker` file and registration for the Ember app assets (cache-first) and the Sockethub assets (network-first). Configuration is in `config/environment.js`.

We'll need our own SW code for Push notifications and more sophisticated cash handling (e.g. for Sockethub assets), of course, so this is meant mostly for loading time improvements in production for now.

<del>Caveat: This branch only works with https://github.com/sockethub/sockethub/pull/204 because the current Sockethub master doesn't add CORS headers to the static asset responses.</del> PR has been merged. So latest `master` should work.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -3,6 +3,7 @@
     "document",
     "window",
     "-Promise",
+    "Promise",
     "linkifyStr"
   ],
   "browser": true,
--- a/config/environment.js
+++ b/config/environment.js
@@ -13,7 +13,7 @@ module.exports = function(environment) {
       }
     },
 
-    sockethubURL: 'localhost:10550',
+    sockethubURL: 'http://localhost:10550',
 
     APP: {
       // Here you can pass flags/options to your application instance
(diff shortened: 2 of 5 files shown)


skddc:
By the way, I think we should set the `enabled` flag to `false` by default in development before merging. Otherwise this incurs an additional page load and cache deletion/creation for every (live)reload.

galfert:
Works as advertised.

But when trying to disable it in development [via the config](https://github.com/67P/hyperchannel/pull/79/files#diff-695693c5dabaffb064bde5b060195ff7R54), I'm seeing this error in the console:

```
A bad HTTP response code (404) was received when fetching the script.
Failed to load resource: net::ERR_INVALID_RESPONSE
Error registering service worker:TypeError: Failed to register a ServiceWorker: A bad HTTP response code (404) was received when fetching the script.
```

Happens even after unregistering the ServiceWorker and deleting all local caches.

Any idea why that is?

skddc:
I'd guess that `includeRegistration` might not care about the other flag. Have you checked `index.html`?

skddc:
Unfortunately, this currently breaks the production build: https://github.com/jkleinsc/broccoli-serviceworker/issues/45

skddc:
Fixed. \o/

PR #85, opened 2017-03-20
Load logs from previous days

### LINKED ISSUE #83: Load logs for previous days
Currently, when joining a channel, we try to load today's logs of that channel.

Would be nice to also provide a button or something to load any earlier messages.

galfert:
... and add date headlines to the channel view.

Closes #83


skddc:
So much better now! Makes Hyperchannel much more useful than before imo.

bluesbrothers

PR #86, opened 2017-03-24
Don't try to load archive when joining unlogged channel

galfert:
Completely missed when implementing #83 that, now that we have a list of logged channels, we don't need to try to load any archive when joining an unlogged channel :)


skddc:
Coolio.

PR #87, opened 2017-03-25
Fix persisting of joined channels in remoteStorage

galfert:
The channels were only persisted when joining via the `/join` command, but not when using the + from the menu. I moved the persisting to the method that both the features use.

Also, when leaving channels, the two properties `channels` and `channelList` of the `Space` model diverted. This prevented the channels from being removed from remoteStorage. I changed the list of channel names to be a computed property based on the actual list of channels instead.

### DIFF
--- a/app/services/smt.js
+++ b/app/services/smt.js
@@ -35,10 +35,10 @@ export default Service.extend({
       rs.kosmos.spaces.getAll().then(spaceData => {
         if (isEmpty(Object.keys(spaceData))) {
           Logger.debug('No space data found in RS. Adding default space...');
-          this.get('storage').addDefaultSpace().then((space) => {
-            this.connectToIRCServer(space);
-            this.get('spaces').pushObject(space);
-            this.instantiateChannels();
+          this.get('storage').addDefaultSpace().then((data) => {
+            this.connectToIRCServer(data.space);
+            this.get('spaces').pushObject(data.space);
+            this.instantiateChannels(data.space, data.channels);
             resolve();
           });
         } else {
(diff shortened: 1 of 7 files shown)


skddc:
w00t! Works like a charm.

Just one minor question about removing a debug log.

skddc:
> I moved the persisting to the method that both the features use

I think this is actually a bad idea (not the concept but the method), because now it's saving spaces multiple times almost at once, when just joining channels at startup. It should only update the space RS document, when the user joins a new channel by user action (that is `/join` or `+`, but not when calling the join method).

app/services/smt.js
```diff
@@ -410,9 +407,7 @@ export default Service.extend({
 
     space.get('channels').removeObject(channel);
 
-    // Update persisted channel list
-    this.get('storage.rs').kosmos.spaces.store(space.serialize())
-      .then(() => this.log('leave', 'stored space', space.get('name')));
```
galfert: It's pretty much the same log message as the `saveSpace` method of the `storage` service logs as well. And the rest of the code is identical as well, that's why I replaced it.

galfert:
I changed the code to only persist the channels when actually joining or leaving a channel.

skddc:
Checked again. Now (or maybe also before) the `/join` command is not cleared from the input field when hitting enter and joining the channel.

galfert:
> the /join command is not cleared from the input field when hitting enter and joining the channel.

I hadn't noticed that. Will check if it happens for me as well.

galfert:
That was indeed an error I introduced with my last commit. It's fixed now.

Thanks for pointing it out.

PR #88, opened 2017-03-29
Add noopener and noreferrer to links with target _blank

galfert:
I saw a talk about Ember security yesterday. One of the suggestions was to always use `noopener` and `noreferrer` as values for the `rel` attribute of links with target `_blank` to prevent a possible XSS vulnerability.

### DIFF
--- a/app/components/message-chat/component.js
+++ b/app/components/message-chat/component.js
@@ -28,14 +28,14 @@ export default Ember.Component.extend({
 
     // Images
     if (content.match(/^(http)s?:\/\/.*(gif|png|jpg|jpeg)$/i)) {
-      out = `<br><a href="${content}" target="_blank" rel="nofollow">` +
+      out = `<br><a href="${content}" target="_blank" rel="nofollow noopener noreferrer">` +
             `<img src="${content}" class="from-image-url" alt="${content}">` +
             `</a>`;
     }
     // Other links
     else {
       out = linkifyStr(content, {
-        linkAttributes: { rel: 'nofollow' },
+        linkAttributes: { rel: 'nofollow noopener noreferrer' },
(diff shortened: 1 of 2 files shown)


skddc:
What's the vulnerability? Wouldn't this also prevent people from seeing Kosmos/Hyperchannel as legitimate referrer in their logs?

galfert:
This gives a good explanation of the possibilities: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

galfert:
Wow, the `noopener` attribute also seems to have performance benefits: https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

skddc:
Well, this basically just means that you could redirect someone and hope they don't notice. And Google also think they shouldn't do anything about it: https://sites.google.com/site/bughunteruniversity/nonvuln/phishing-with-window-opener

skddc:
... if the main benefits come from `noopener` and the main drawback from the FF "workaround" of also adding `noreferrer`, then we could e.g. only add the former as the best tradeoff.

skddc:
Yup, looks like it would hide the referrer entirely: https://www.w3.org/TR/html5/links.html#link-type-noreferrer

gregkare:
Looks like noopener is implemented in Firefox 52: https://bugzilla.mozilla.org/show_bug.cgi?id=1222516

skddc:
Great!

@galfert Wanna change it to just do that then?

galfert:
Ah, cool. Thanks for the info, @gregkare. I will remove the `noreferrer` then.

galfert:
I removed the `noreferrer` value.

PR #90, opened 2017-04-03
Fix archive loading for today's logs

### LINKED ISSUE #89: Hyperchannel loads logs from previous day, but not current one
Currently, on app startup, Hyperchannel will load logs of the previous day (with content), but not the current one. Not sure if it's always or in certain situations, but I could reproduce it in the middle of (GMT) day today, so it doesn't seem to be some local-timezone issue or similar.

galfert:
Paired with @skddc on this.

Fixes #89

Rename some variables to make a little more sense and use UTC dates everywhere.

### DIFF
--- a/app/components/channel-container/component.js
+++ b/app/components/channel-container/component.js
@@ -49,7 +49,7 @@ export default Component.extend({
       this.get('smt').loadLastMessages(
         this.get('channel.space'),
         this.get('channel'),
-        this.get('channel.previousLogsDate')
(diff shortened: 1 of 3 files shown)


skddc:
Unfortunately, now it's skipping one day when going back from today's history. Looks like we still need to find out why the actual skip happened within UTC time.

galfert:
Dang :(

Will try to have another look at it later.

galfert:
Ok, now it should finally all work.

skddc:
Yup, all looking great now.

PR #94, opened 2017-04-11
Refactoring (+ space settings/presets)

skddc:
When I intended to continue hacking on the settings branch yesterday, it got a little bit out of hand and I ended up refactoring most of the `smt` service and splitting out IRC functions to its own `sockethub-irc` service in preparation for a `sockethub-xmpp` and potential other adapters. In theory the AS objects should look mostly the same in both IRC and XMPP, but this way it's much easier to hack forward fast, and then later compare which parts can be unified, without breaking either adapter while hacking on the other one.

Anyway, this doesn't introduce major changes, so maybe just scan the code and smoke-test if stuff still works. I actually found a bug which entered `master` recently, and also fixed that just now.

Edit: turns out we were still testing on node 4, and that didn't likesome new syntax outside of strict mode. So I also changed it to test node 6 and require that for devs, too.

### DIFF
--- a/app/services/sockethub-irc.js
+++ b/app/services/sockethub-irc.js
@@ -17,6 +17,15 @@ export default Ember.Service.extend({
   logger: service(),
   coms: service(),
 
+  /**
+   * @public
+   * @method connect
+   *
+   * - Creates an ActivityStreams person object for
+   *   future use
+   * - Emits credentials for future IRC server messages,
+   *   like e.g. `join`
+   */
   connect(space) {
     this.sockethub.ActivityStreams.Object.create({
       '@id': space.get('sockethubPersonId'),
--- a/app/services/sockethub-irc.js
+++ b/app/services/sockethub-irc.js
@@ -1,9 +1,3 @@
-/**
- * @module sockethub-irc
- *
- * This service provides helpers for SocketHub IRC communications
- */
-
 import Ember from 'ember';
 
 const {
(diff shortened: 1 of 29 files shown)


tests/unit/services/sockethub-irc-test.js
```diff
@@ -0,0 +1,17 @@
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('service:sockethub-irc', 'Unit | Service | sockethub irc', {
+  needs: [
+    'service:ajax',
+    'service:logger',
+    'service:remotestorage',
+    'service:coms'
```
galfert: See comment above, `coms` is not needed.

app/services/sockethub-irc.js
```diff
@@ -0,0 +1,180 @@
...
+} = Ember;
+
+/**
+ * This service provides helpers for SocketHub IRC communications
+ * @module hyperchannel/services/sockethub-irc
+ */
+export default Ember.Service.extend({
+
+  logger: service(),
+  coms: service(),
```
skddc: This service's unit test actually makes sure it has access to coms. When we move more functions here it'll be needed I think.

galfert:
Did some smoke testing and couldn't find any problems.

skddc:
Cool, thanks.

galfert:
Btw, I started to add modals, so we can use them for the preset config. Will try to finish that tomorrow.

skddc:
Oh, nice!

PR #95, opened 2017-04-14
Space settings modal

galfert:
This adds a modal for the space settings presets.

The same modal could also be used for creating a new space without preset or editing existing spaces. But this has not been done in this PR.


skddc:
firsttry

PR #97, opened 2017-04-21
Refactorings

galfert:
This fixes some eslint warnings and removes code duplication, hopefully pleasing the code climate gods.


skddc:
LGTM

Issue #12, opened 2015-05-15
Support colored messages


skddc:
E.g. for GitHub notifications.

skddc:
Basic support is in!

skddc:
Left to do: bold, underline, italic. Maybe, but I don't find it important: reverse, clear colors.

PR #101, opened 2017-05-03
Add support for bold and italic messages

gregkare:
Connected to #12 

I have used spans with classes, I'm open to other ways to do it though

### DIFF
--- a/app/components/message-chat/component.js
+++ b/app/components/message-chat/component.js
@@ -46,9 +46,11 @@ export default Ember.Component.extend({
     }
 
     // Colors
-    out = out.replace(/\u0003(\d+)/g,'<span class="color-$1">')
-             .replace(/\"color-(\d)\"/g,'"color-0$1"')
-             .replace(/\u000f/, '</span>');
+    out = out.replace(/\u0003(\d+)/g, '<span class="color-$1">')
+             .replace(/\"color-(\d)\"/g, '"color-0$1"')
+             .replace(/\u0002/g, '<span class="bold">')
+             .replace(/\u001D/g, '<span class="italic">')
+             .replace(/\u000f/g,  '</span>');
 
     return htmlSafe(out);
   })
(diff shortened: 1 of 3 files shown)


skddc:
Cool. I think we should use `<em>` for italic and `<strong>` for bold instead. Those are the semantic HTML equivalents to the classes.

gregkare:
Thanks, changing it

gregkare:
I should have tried it, I used spans because I thought the `<em>` and `<strong>` tags would just be open and I'd have to close them manually :)

gregkare:
Adding the semantic tags and trying to close them using replacement didn't work at all, the results were super unpredictable.  In a342c16 I have switched to an npm module to do the actual replacements, dealing with nesting and resetting properly. I couldn't find a way to use `<strong>` and `<em>` tags instead of `<span>` (because Handlebars cannot do comparisons like equality without helpers, and I couldn't make any of the solutions work)

Now it supports bold, italic, underlined and proper nesting:

screen shot 2017-05-04 at 15 20 20

skddc:
As discussed yesterday, we don't actually want underlines to be used with something that's not a link, and we add our own underlines when turning URLs into actual links in the message component.

This worked almost with our own implementation that was extremely simple and didn't rely on a 3rd-party module (which will have to be included in the app code). I think we should go back to our solution, use spans instead of semantic elements for now, and remove underline support. No external dependencies, no additional code bloat and supports all the features we need.

skddc:
Looks like @gregkare missed my last comment, but he's on vacation for a week from today on.

I'll check if I can revert to an earlier commit and just remove the underline.

skddc:
@gregkare Wanna have a look at this again? Seems like we just need to pick a previous commit.

gregkare:
Ack, will do!

gregkare:
I have pushed the simplest change to support bold and italic and removed underlined support. It still has the issue of the style reset only resetting one tag, but I don't see a way to do this properly without using a real IRC style parser, so this is good enough

skddc:
Great!

Issue #100, opened 2017-05-03
Leave channels via UI (instead of command)


skddc:
Currently, one needs to do write a `leave` or `part` command in order to leave a channel. Non-technical users should likely have a UI-element available for that.

PR #104, opened 2017-05-04
Leave channel via UI

gregkare:
Connected to #100


gregkare:
This is working! I have added basic styling. There's an error message when closing the last channel, but that appears to be in master as well when using `/part` or `/leave`

skddc:
This was actually almost entirely a UX question. Could you add a screenshot and start a conversation about the design?

gregkare:
Here's a screenshot of what I've done:

leave

I haven't added a confirmation prompt, clicking the x leaves the channel/query

skddc:
Personally, I think that's not a great place, because it's easy to accidentally hit those, and also it just doesn't look nice to have the channel list plastered with leave buttons.

belenbarrospena:
Since it looks like one channel is always selected, what about a button at the bottom of the left panel to leave the current channel? 

It creates less visual noise than having a button for each channel always displayed, and it's out of the way enough that you hopefully won't hit it by mistake. 

It is just a suggestion: take it with a pinch of salt, since I know nothing about what you are building :)

button_purple
button-blue

skddc:
@belenbarrospena Thanks for jumping in (it looks like you even set up Hyperchannel/Sockethub on your machine?).

I would add to this, that there are more channel actions to consider, so grouping them like Mattermost does could make sense (it's the easy way out of course, but not necessarily the most elegant imo):

screenshot from 2017-05-05 13-11-50

Changing the channel topic would be an example for such an item, that we need to add in the near future.

belenbarrospena:
> @belenbarrospena Thanks for jumping in (it looks like you even set up Hyperchannel/Sockethub on your machine?).

No, I didn't: I only visited https://hyperchannel.kosmos.org :)

> I would add to this, that there are more channel actions to consider, so grouping them like Mattermost does could make sense (it's the easy way out of course, but not necessarily the most elegant imo):

This is good context: thank you! The Mattermost design is perfectly good: my only issue with it is that I wouldn't pack "leave" with the rest of the options in there, which are more "managerial" in nature, for lack of a better word. 

You could add them to the "settings" section in the right panel, which now seems to include application-wide things, but you could split it and have a section for channel-related actions.

Or you could keep them on the left panel by adding a button next to the selected channel. That would probably require giving that panel, and each of the channels, a bit more room to accommodate the extra options. 

Or you could add a menu to the top bar in the central panel, where you currently display the channel name and description. You already have 2 options in there right-aligned: you could have an extra one. 

So there are tons of alternatives, but to summarise, I would not necessarily pack the "leave" channel action with channel management options, since they seem of a different nature. 

Just my 2c :)

Cheers

belenbarrospena:
funnily enough, Slack seems to do the same thing: it packs the 'leave' channel options with the channel settings. I am still not convinced! ;)

leave-with-settings

skddc:
> to summarise, I would not necessarily pack the "leave" channel action with channel management options, since they seem of a different nature.

Good point. I agree.

The big difference between Mattermost/Slack and Kosmos/Hyperchannel is that you can join channels much more easily with Kosmos (public IRC/XMPP channels on e.g. Freenode), so it's also more common to leave (and later re-enter) channels. Which makes it even less of a "managerial/administrative" action.

silverbucket:
Thanks for the feedback @belenbarrospena 

I like the idea of the leave channel in the lower left corner, and think it would be nice to avoid right-click functionality if possible.

For example, the topic could be set simply by clicking in the topic area and getting a textarea box to submit a new topic. This would only switch to text input if you have the right permissions.

What other channel-specific operations would you need to do that couldn't be expressed elsewhere in the UI, rather than with a right-click?

galfert:
> There's an error message when closing the last channel, but that appears to be in master as well when using /part or /leave

FYI: I fixed that in https://github.com/67P/hyperchannel/pull/112

galfert:
When I leave a channel that is not currently active, I'm still switched to the last channel listed.

skddc:
Good point. In that case it should just keep you in the current channel I suppose.

skddc:
@galfert I changed it so it doesn't switch channels when closing one that isn't open. I also found a small bug, in that it would select the last channel from the unsorted channel list, thus jumping to a channel that wasn't last.

Issue #21, opened 2015-05-15
Improve op user rendering in member list


raucao:
Currently, they just have an @ in front of their name, because that's plain IRC. Should remove the @ and add a status to the user object, as well as highlight it visually somehow.

gregkare:
It looks like there's a bug in the current master when trying to message an op by clicking their name in the channel list. Since the username is prefixed with @ it does not work:

```
  sockethub-platform-irc Unprocessed message [irc://hyper_greg@irc.freenode.net]:  IRCMessage {
  prefix: 'cherryh.freenode.net',
  command: 'ERR_NOSUCHCHANNEL',
  params: [ 'hyper_greg', '@ChanServ', 'No such channel' ],
  server: 'cherryh.freenode.net',
  raw: ':cherryh.freenode.net 403 hyper_greg @ChanServ :No such channel',
```

skddc:
@gregkare That would be a different issue. Care to create one specifically for that bug?

gregkare:
Will do!

PR #106, opened 2017-05-04
Add a class to ops in the users list

gregkare:
Refs #21 

It works but I haven't added styling yet. This should also fix messaging ops from the user list


skddc:
I [increased](https://github.com/67P/hyperchannel/commit/40553ae5ed59b6e9c4658c68435f768f2dbf3677) the mass threshold for the CodeClimate duplication warnings. These should be green in the future, unless you have actual duplication going on.

gregkare:
I'm going to remove the User model for now, switch back the user list to an array of strings and use a helper to render each user list. The sort order works automatically because of the prefixes

gregkare:
This is ready to review. I haven't changed the UI for it, but this adds a "normal" class to normal users and "op" to ops

app/components/link-to-username/component.js
```diff
@@ -0,0 +1,25 @@
...
+  computed
+} = Ember;
+
+export default Ember.Component.extend({
+  username: null,
+
+  role: computed('username', function() {
+    if (this.get('username').startsWith('@')) {
+      return 'op';
+    } else {
```
skddc: Shouldn't we have an `else if` for moderators? You can see them e.g. in the #indieweb room Freenode I believe.

app/components/link-to-username/component.js
```diff
@@ -0,0 +1,25 @@
...
+      return 'normal';
+    }
+  }),
+
+  usernameWithoutPrefix: computed('username', function() {
+    if (this.get('username').startsWith('@')) {
+      return this.get('username').replace('@', '');
```
skddc: Same as the other question: what about other known prefixes?

app/components/link-to-username/component.js
gregkare: `+` is voice, not moderator (meaning you can talk when the channel is set to silent, when only voices and ops can talk). `%` is half op, which is close to moderator (they can voice and kick/ban people with a lower level than themselves.

I'm adding support for voice and half op

skddc: Cool.

gregkare:
`+` comes before `@` when doing a `.sort()`, so we'll need a custom sort function

skddc:
Sounds like a seperate new issue to me, because they're already sorted by that first character.

gregkare:
There's still an issue with this PR: there's a double rendering issue.

@galfert Do you know what I should do to fix it?

```
"Assertion Failed: You modified "unreadMessagesClass" twice on <(unknown mixin):ember1638> in a single render. It was rendered in "component:channel-nav" and modified in "component:link-to". This was unreliable and slow in Ember 1.x and is no longer supported. See https://github.com/emberjs/ember.js/issues/13948 for more details.
EmberError@http://localhost:4200/assets/vendor.js:22898:19
```

This seems to happen when joining a new channel, and goes away when reloading Hyperchannel

skddc:
@galfert Did you see @gregkare's question here?

galfert:
No, I missed that. Might be some runloop issue. I'll have a look.

galfert:
When testing this I had the (already known and fixed) errors due to the incompatibility with the recent Sockethub changes. So I merged in the current master. But I still couldn't reproduce the error you mentioned.

The feature itself worked. But I took the liberty to refactor it a little bit to be less "if-elsy" :)
I also added some unit tests.

Overall I think it's good to merge now.

tests/unit/components/link-to-username-test.js
```diff
@@ -0,0 +1,41 @@
...
+  test
+} from 'ember-qunit';
+
+moduleForComponent('link-to-username', {
+  // Specify the other units that are required for this test
+  // needs: ['component:foo', 'helper:bar']
+  unit: true
+});
+
+test('role returns the IRC role of the user', function(assert) {
```
skddc: Does it really only do it on IRC though? ;)

galfert: Good point. It actually does it for all users.

skddc: I did a quick check with XMPP and it would be possible to impersonate mods and ops there, as you can choose any nickname you want, incl. special characters:

screenshot from 2017-05-11 14-49-11

skddc: @galfert Should we merge this, so it's ready to be styled tho? We could create a seperate issue for making this feature IRC-only (or rather: add the tag based on the protocol's API).

galfert: Yes, let's merge and create a separate issue.

PR #109, opened 2017-05-05
Change username when getting update event from Sockethub

### LINKED ISSUE #108: Handle username updates
When a username is already taken, Sockethub sends an `update` message with a new username that was assigned instead.

galfert:
Closes #108

### DIFF
--- a/app/models/space.js
+++ b/app/models/space.js
@@ -22,9 +22,14 @@ export default Ember.Object.extend({
   },
   channels   : null, // Channel instances
 
+  // Keep a list of all old sockethubPersonIds, because there might
+  // still be coming events from Sockethub for those.
+  previousSockethubPersonIds: null,
+
   init() {
     this._super(...arguments);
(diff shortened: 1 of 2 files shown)


skddc:
Weird, for me it uses the protected username now (instead of the new one):

screenshot from 2017-05-08 17-58-59

galfert:
Weird.

I can't test this right now anymore with my current Sockethub version. Would be great if we could merge #113 first, then I will rebase this PR on master and see what's up.

galfert:
Judging from the screenshot I think you weren't really connected, because I don't see those messages you wrote in my logs. My guess is that was due to the Sockethub issues/incompatibilities we fixed today.

I updated this branch with the recent changes in master and for me the name change worked as expected just now.

skddc:
Ok, makes sense. Thanks. Trying it again.

skddc:
w00t!

Issue #41, opened 2016-10-10
Support XMPP MUC


skddc:
5apps will start dogfooding Kosmos for our private company chats within the next 4 weeks. Ideally, we'd like to use XMPP instead of IRC from the start.

[I opened an issue](https://github.com/sockethub/sockethub/issues/197) to re-create the Sockethub platform, but I'm not actually sure that's the best way, because with BOSH we could also forgo the whole Sockethub dependency and connect to the XMPP server directly.

skddc:
New milestone for the SH platform: https://github.com/sockethub/sockethub-platform-xmpp/milestone/1

PR #112, opened 2017-05-05
Add function which processes channel joins, including failures.

silverbucket:
Connected to #41

### DIFF
--- a/app/controllers/space/base_channel.js
+++ b/app/controllers/space/base_channel.js
@@ -1,12 +1,22 @@
 import Ember from 'ember';
 import Message from 'hyperchannel/models/message';
 
-export default Ember.Controller.extend({
+const {
+  Controller,
+  inject: {
+    controller,
+    service
+  },
+  Logger,
+  isPresent
+} = Ember;
+
+export default Controller.extend({
 
   newMessage: null,
-  space: Ember.inject.controller(),
-  coms: Ember.inject.service(),
-  storage: Ember.inject.service('remotestorage'),
+  space: controller(),
+  coms: service(),
+  storage: service('remotestorage'),
 
   createMessage(message, type) {
     return Message.create({
@@ -48,7 +58,7 @@ export default Ember.Controller.extend({
       if (availableCommands.includes(command.toLowerCase())) {
         this.send(command + 'Command', commandSplitted.slice(1));
       } else {
-        Ember.Logger.warn('[channel]', 'Unknown command', commandText);
+        Logger.warn('[channel]', 'Unknown command', commandText);
       }
 
       this.set('newMessage', null);
@@ -66,7 +76,11 @@ export default Ember.Controller.extend({
       let channelName = this.get('model.name');
       this.get('coms').removeChannel(space, channelName);
       let lastChannel = space.get('channels.lastObject');
-      this.transitionToRoute('space.channel', space, lastChannel);
+      if (isPresent(lastChannel)) {
+        this.transitionToRoute('space.channel', space, lastChannel);
+      } else {
+        this.transitionToRoute('space', space);
+      }
     },
 
     leaveCommand: function() {
--- a/app/controllers/space/base_channel.js
+++ b/app/controllers/space/base_channel.js
@@ -30,7 +30,7 @@ export default Controller.extend({
 
   actions: {
     sendMessage: function() {
-      assert('Please define "sendMessage"');
+      assert('Please define "sendMessage" in the controller inheriting from BaseChannel');
     },
 
     executeCommand: function(newMessage) {
--- a/app/controllers/space/base_channel.js
+++ b/app/controllers/space/base_channel.js
@@ -8,8 +8,7 @@ const {
     service
   },
   Logger,
-  isPresent,
-  assert
+  isPresent
 } = Ember;
 
 export default Controller.extend({
@@ -29,8 +28,17 @@ export default Controller.extend({
   },
 
   actions: {
-    sendMessage: function() {
-      assert('Please define "sendMessage" in the controller inheriting from BaseChannel');
+    sendMessage: function(newMessage) {
+      let message = this.createMessage(newMessage, 'message-chat');
+
+      this.get('coms').transferMessage(
+        this.get('space.model'),
+        this.get('model'),
+        message.get('content')
+      );
+
+      this.get('model.messages').pushObject(message);
+      this.set('newMessage', null);
     },
 
     executeCommand: function(newMessage) {
--- a/app/controllers/space/base_channel.js
+++ b/app/controllers/space/base_channel.js
@@ -8,7 +8,8 @@ const {
     service
   },
   Logger,
-  isPresent
+  isPresent,
+  assert
 } = Ember;
 
 export default Controller.extend({
@@ -28,17 +29,8 @@ export default Controller.extend({
   },
 
   actions: {
-    sendMessage: function(newMessage) {
-      let message = this.createMessage(newMessage, 'message-chat');
-
-      this.get('coms').transferMessage(
-        this.get('space.model'),
-        this.get('model.sockethubChannelId'),
-        message.get('content')
-      );
-
-      this.get('model.messages').pushObject(message);
-      this.set('newMessage', null);
+    sendMessage: function() {
+      assert('Please define "sendMessage"');
     },
 
     executeCommand: function(newMessage) {
--- a/app/controllers/space/channel.js
+++ b/app/controllers/space/channel.js
@@ -1,4 +1,23 @@
 import BaseChannel from 'hyperchannel/controllers/space/base_channel';
 
 export default BaseChannel.extend({
+
+  actions: {
+    sendMessage: function(newMessage) {
+      let message = this.createMessage(newMessage, 'message-chat');
+
+      this.get('coms').transferMessage(
+        this.get('space.model'),
+        {
+          '@id': this.get('model.sockethubChannelId'),
+          '@type': 'room'
+        },
+        message.get('content')
+      );
+
+      this.get('model.messages').pushObject(message);
+      this.set('newMessage', null);
+    }
+  }
+
 });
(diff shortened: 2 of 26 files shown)


silverbucket:
This successfully handles JOIN failures, however since the space is looked up by hostname, and the hostname for the XMPP JID is mixed up (in our test case, the user is `user@host.com` while the server is `xmpp.host.com`, therefore it's unclear what is actually the JID and/or what is the official way of referencing an account.

If, as we discussed today, the JID is `user@host.com/resource` then it does not give enough information about where this user exists, it would seem like this means it can't be the read JID. 

So, is the real JID `user@host.com@xmpp.host.com/resource` ? There are so many pieces involved right now its a bit hard to work through this alone. At any rate, this pull request is functional except for one piece @galfert said he'd look into which is that the personId lookup is failing due to an incorrect server name (for xmpp)

gregkare:
As discussed on Skype the JID is `user@host.com/resource` even if the server has DNS entries set up to announce to clients that the server is running on another host (xmpp.host.com for example)

galfert:
I pushed some changes so at least IRC works again in this PR. Haven't gotten to trying XMPP yet. Will do that later.

galfert:
Update: I fixed some problems and bugs related to XMPP, and also refactored some of the code along the way. But right now I'm at a point where I just don't get any feedback from Sockethub when trying to join a channel, except for the `completed` events. But no `message` or `failure` events.

And I also can't see anything of relevance in the Sockethub logs itself.

I am connected to the server though, as I can see the presence updates of the other users on the server.

@silverbucket: do you maybe have some time on friday to further pair on this a little bit?

config/space-presets.json
```diff
@@ -48,7 +48,7 @@
     }
   },
   {
-    "id": "5apps",
+    "id": "xmpp-5apps",
```
skddc: This was actually intentional, because the idea is that you can change the type of the same space when migrating, without moving everything.

galfert: But then it conflicts with the [computed `id`](https://github.com/67P/hyperchannel/pull/112/files#diff-be00797e644f8bf6fa99549fc3b20ffcR49) which is based on the name and is used for the URLs.

skddc: What exactly is the conflict? I don't see what that's doing.

galfert: Ember is using the id to generate the URLs. I don't know exactly what was going on in detail, but when the preset uses "5apps" as id, it links channels to `/5apps/channel/<channelslug>` but then would error when trying to navigate to it, because it wouldn't find the space by that id.

Using the id based on the name fixes that.

skddc: Sorry, I still don't understand what that means really. Why should that not work?

skddc: ... in fact, that's *exactly* the URL I wanted it to use.

galfert: I changed the code to not use a computed property for the id anymore. Now it's only generating an id from the name when no explicit id was given. This way everything works fine now.

silverbucket:
@galfert sorry I didn't notice you ping me to pair. I'm free this weekend if you are.

galfert:
I finally managed to successfully join XMPP MUC rooms. The problem was that that you have to add your nickname as a resource to the room id, e.g. `room@muc.host.com/nickname`.

Messaging in rooms work now. But it's still missing the list of users in the room.

As for direct messages, we can receive them but the outgoing messages don't seem to be sent to the recipient yet. I couldn't see any obvious error messages about that in the Sockethub logs, but I also didn't look into that very thoroughly yet.

app/controllers/space/base_channel.js
```diff
@@ -28,17 +29,8 @@ export default Controller.extend({
...
-      this.get('coms').transferMessage(
-        this.get('space.model'),
-        this.get('model.sockethubChannelId'),
-        message.get('content')
-      );
-
-      this.get('model.messages').pushObject(message);
-      this.set('newMessage', null);
+    sendMessage: function() {
+      assert('Please define "sendMessage"');
```
skddc: Huh?

app/controllers/space/channel.js
```diff
@@ -1,4 +1,23 @@
...
+  actions: {
+    sendMessage: function(newMessage) {
+      let message = this.createMessage(newMessage, 'message-chat');
+
+      this.get('coms').transferMessage(
+        this.get('space.model'),
+        {
+          '@id': this.get('model.sockethubChannelId'),
+          '@type': 'room'
+        },
```
skddc: We shouldn't deal with AS manually in arguments to `coms` methods, no?

app/controllers/space/user_channel.js
```diff
@@ -1,4 +1,23 @@
...
+  actions: {
+    sendMessage: function(newMessage) {
+      let message = this.createMessage(newMessage, 'message-chat');
+
+      this.get('coms').transferMessage(
+        this.get('space.model'),
+        {
+          '@id': this.get('model.sockethubChannelId'),
+          '@type': 'person'
+        },
```
skddc: Same as above, except with the added argument that this is duplicating code.

app/services/sockethub-xmpp.js
```diff
@@ -136,6 +141,64 @@ export default Ember.Service.extend({
...
+      type: messageType,
+      date: new Date(message.published),
+      nickname: message.actor.displayName || message.actor['@id'],
+      content: message.object.content
+    });
+
+    // TODO should check for message and update sent status if exists
+    if (channelMessage.get('nickname') !== space.get('userNickname')) {
+      channel.addMessage(channelMessage);
+    }
```
skddc: Apart from this method being super long, it seems like considerable code duplication to me.

skddc:
> The problem was that that you have to add your nickname as a resource to the room

We shouldn't have to do that in Hyperchannel! The whole point of Sockethub platforms is to abstract that away from apps.

skddc:
I can see tons of changes, and one of them is speeding up the test suite. However, there's not a single new or changed test in here?

app/controllers/space/base_channel.js
galfert: This is just a base class. The classes that inherit from this (e.g. `Channel` and `UserChannel` ) need to implement the actual method. This throws an error with the hint in case you inherit without defining the method yourself.

galfert:
The recent changes are just me trying to get everything to work at first. It's not refactored, cleaned up or tested yet.

app/controllers/space/base_channel.js
skddc: Well, I didn't understand the hint here, so maybe it should say what you just explained. ;)

skddc:
Got it. Only saw the WIP after my last comment earlier. ;)

app/controllers/space/base_channel.js
galfert: Done

app/models/base_channel.js
```diff
@@ -80,7 +81,9 @@ export default Ember.Object.extend({
   },
 
   addUser(username) {
-    this.get('userList').pushObject(username);
+    if (isEmpty(this.get('userList').find(name => name === username))) {
```
skddc: You can actually write this as `.findBy('name', username)` I believe.

app/services/coms.js
```diff
@@ -448,7 +448,7 @@ export default Service.extend({
             this.updateUsername(message);
             break;
           case 'presence':
-            Logger.debug('Presence update:', message.actor['@id'], message.object.presence, message.object.status);
+            this.getSockethubPlatformFor(message.context).handlePresenceUpdate(message);
```
skddc: Why not just use `this.get(message.context).handlePresenceUpdate()`? Also, isn't this XMPP-specific?

app/models/base_channel.js
galfert: No, the `userList` is just an array.

skddc: Even better. Then we could do [`if (this.get('userList').includes(username))`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes?v=control).

galfert: doh m(

:)

galfert:
I'm still writing some tests for the new methods. But other than that this should be ready for you to have a look and give it a try.

skddc:
@galfert It's all red. Should I wait maybe?

galfert:
Added a bunch of tests as well now. This should be all ready to go now.

skddc:
There I fixed it. Pushing from the TXL runway, so didn't test in browser anymore, but trusting the specs. :)

skddc:
@galfert Ok, so is there anything missing now?

To confirm: it's correct that the XMPP join, which under the hood is actually sending presence to a channel, is not resulting in my user being shown in the channel list currently, right?

Also, sending messages is not included here on purpose, correct?

galfert:
> [...]  is not resulting in my user being shown in the channel list currently, right?

Why would you show up in the channel list? When joining a channel, you should be seen in the user list of that channel. In Hyperchannel, as well as in other clients.
When you receive a direct message, the other user's JID should show up as a channel in the channel list.

> Also, sending messages is not included here on purpose, correct?

Not sure I understand. You can send and receive messages in channels. What doesn't work yet is initiating direct message chats with a user by clicking the user in the user list of a channel. But when the other user sends you a direct message first, you can also send direct messages to that user.

skddc:
> Why would you show up in the channel list? When joining a channel, you should be seen in the user list of that channel. In Hyperchannel, as well as in other clients.

That's what I meant. But it didn't work.

> Not sure I understand. You can send and receive messages in channels.

It didn't work, so I assumed it's not implemented yet.

skddc:
For posterity, I wrote this on IRC after we seemed to have found the bug, but it was actually a different one:

```
<raucao> galfert: i'm back to it not working, but without the replace-connection error
<raucao> but also with the same procedure (wrong jid first, then correct one)
<raucao> all the sockethub messages get completed
<raucao> that's what i had before (and i was wondering why there's an error now, which i would have reported then)
```

PR #116, opened 2017-05-08
Use postcss and autoprefixer, remove old bourbon mixins

### LINKED ISSUE #114: Use CSS autoprefixer
Right now, when starting the the app, Bourbon is logging a lot of deprecation messages, like

> WARNING: [Bourbon] [Deprecation] `flex` is deprecated and will be removed in 5.0.0. We suggest using an automated prefixing tool, like Autoprefixer.

Possible Ember addon to use: [ember-cli-postcss](https://github.com/jeffjewiss/ember-cli-postcss)

skddc:
This switches some of Bourbon's deprecated mixins to normal syntax, for
which prefixes are added by postcss/autoprefixer instead.

connected to #114
closes #114

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -4,7 +4,8 @@
     "window",
     "-Promise",
     "Promise",
-    "linkifyStr"
+    "linkifyStr",
+    "process"
   ],
   "browser": true,
   "boss": true,
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -1,20 +1,20 @@
 @mixin app-column() {
-  @include display(flex);
-  @include align-items(stretch);
-  @include flex-direction(column);
-  @include justify-content(flex-start);
+  display: flex;
+  align-items: stretch;
+  flex-direction: column;
+  justify-content: flex-start;
 
   header {
-    @include flex(0 0 $headerHeight);
+    flex: 0 0 $headerHeight;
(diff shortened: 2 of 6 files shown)


.jshintrc
```diff
@@ -4,7 +4,8 @@
     "window",
     "-Promise",
     "Promise",
-    "linkifyStr"
+    "linkifyStr",
+    "process"
```
galfert: What is this for? I don't see `process` used anywhere in this PR.

skddc: My JSHint was complaining in `ember-cli-build.js` with the existing code.

galfert:
I can see network requests for http://localhost:4200/assets/vendor.css that fail with a 404 now.

skddc:
Ouch. One would assume the `exclude` for a specific plugin means it's not run against those files, not that those files are then dismissed from the build entirely. I'll check the options again.

skddc:
I tried to switch to `compile` from `filter`, which is supposed to do exactly what we need (just compile `hyperchannel.css`, not `vendor.css`. But it doesn't work for me. See commit message.

skddc:
The author of the plugin fixed the issue I had with the `filter` and `exclude` and published a new version. I reverted to the previous commit and now it works as intended.

Please squash-merge and use the first commit's message.

PR #124, opened 2017-06-19
Mark channels of all spaces as invisible when switching

### LINKED ISSUE #122: Mention highlights not yet working with XMPP channels

galfert:
Fixes #122

When there are multiple spaces, switching channels still kept the channel of the second space marked as visible, preventing the highlight when new messages arrived for that channel.


skddc:
Nice!

PR #128, opened 2017-06-23
Connect IRC channel after getting userlist

### LINKED ISSUE #118: Channel is shown as connected too early
We currently show channels as connected when we receive the `completed` event for the `join` message from Sockethub. But that just indicates that the `join` was sent to IRC now.

When having a different IRC client open at the same time, one can see that the actual moment a user is joining a channel, is the moment Hyperchannel receives the first userlist from Sockethub.

So should we rather use that message to mark the channel as connected?

galfert:
Fixes #118

This is the moment when the user is also shown as online to other clients.

### DIFF
--- a/app/services/sockethub-irc.js
+++ b/app/services/sockethub-irc.js
@@ -91,7 +91,7 @@ export default Ember.Service.extend({
 
   handleJoinCompleted(space, message) {
     var channel = space.get('channels').findBy('sockethubChannelId', message.target['@id']);
-    if (!isEmpty(channel)) {
+    if (channel) {
       this.observeChannel(space, channel);
     }
   },
--- a/app/services/sockethub-irc.js
+++ b/app/services/sockethub-irc.js
@@ -92,7 +92,6 @@ export default Ember.Service.extend({
   handleJoinCompleted(space, message) {
     var channel = space.get('channels').findBy('sockethubChannelId', message.target['@id']);
     if (!isEmpty(channel)) {
-      channel.set('connected', true);
       this.observeChannel(space, channel);
     }
   },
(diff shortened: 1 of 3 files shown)


app/services/sockethub-irc.js
```diff
@@ -92,7 +92,6 @@ export default Ember.Service.extend({
   handleJoinCompleted(space, message) {
     var channel = space.get('channels').findBy('sockethubChannelId', message.target['@id']);
     if (!isEmpty(channel)) {
```
skddc: Shouldn't this be `isPresent` instead (and also be made the same code as above where it's just `is(channel)`)?

tests/unit/services/coms-test.js
```diff
@@ -62,3 +62,47 @@ test('#transferMessage calls transferMessage on the appropriate transport servic
...
+    },
+    "published": "2017-06-23T15:44:54.383Z"
+  };
+
+  const space = Space.create({
+    protocol: 'IRC',
+    name: 'Freenode',
+    server: { hostname: 'irc.freenode.net' }
+  });
+  const channel = Channel.create({
```
skddc: Shouldn't this be `let`, as the object is going to be changed?

app/services/sockethub-irc.js
galfert: true

tests/unit/services/coms-test.js
skddc: Oh wow, totally had the same misconception. Thanks for the link, that's a good post!

skddc:
Works as intended afaics.

Issue #80, opened 2017-02-04
ServiceWorker problem in FF (reloads many times)


skddc:
I just deployed the SW-enabled `master` to https://hyperchannel.kosmos.org. Looks good in Chrome on my phone, but breaks horribly in FF stable on Linux for me. It thinks that the cache is outdated at least once, but sometimes going into an infinite loop, deleting caches and reloading the app endlessly.

gregkare:
I'm getting the same infinite loop on Firefox stable (OS X)

skddc:
I found out what the problem is (roughly): it's a bug in FF, dealing with multiple ServiceWorkers. The 5apps shim worker (which does literally nothing, it's just a shim with no code to enable install dialogs) "conflicts" with the caching worker in this case. Chrome handles it just fine, and I couldn't find anything in the cache worker code that could make it fail.

skddc:
We could check out https://developers.google.com/web/tools/workbox/

PR #129, opened 2017-06-23
Update SW deps, use SW in dev

skddc:
This updates the ServiceWorker-related deps and enables SW in development. It actually works just fine in FF in development, but not with the hosted version. I'll investigate what's wrong on the 5apps side of things, but it doesn't look like a bug in the plugin. Even more so, because nobody else reported it on the plugin source repo.

refs #80


galfert:
Works great.

PR #130, opened 2017-06-26
Handle channel user list for XMPP

### LINKED ISSUE #127: Missing channel userlist when reloading browser
For all channels that the user already joined, after a browser reload the userlist of those channels is empty.

We probably need to do something similar to IRC, where we send an `observe` activity for the channel's user attendance after joining.

galfert:
Closes #127 

This queries for the room attendance list when joining a new room and handles the corresponding response from Sockethub.

It needs the implementation for Sockethub XMPP platform, which I created a PR for at sockethub/sockethub-platform-xmpp#15

Right now this only seems to work properly when connected to XMPP alone, but not when also connected to IRC. I think this is related to the issue reported at sockethub/sockethub-platform-irc#22, in which @skddc already suspected it might not just be an IRC issue but a general multiple connections issue.

### DIFF
--- a/app/services/sockethub-xmpp.js
+++ b/app/services/sockethub-xmpp.js
@@ -94,8 +94,8 @@ export default Ember.Service.extend({
   handleJoinCompleted(space, message) {
     const channelId = message.target['@id'].split('/')[0];
     const channel = space.get('channels').findBy('sockethubChannelId', channelId);
-    if (!isEmpty(channel)) {
-      channel.set('connected', true);
+    if (channel) {
+      this.observeChannel(space, channel);
     } else {
       Logger.warn('Could not find channel for join message', message);
     }
@@ -188,6 +188,26 @@ export default Ember.Service.extend({
     if (channelMessage.get('nickname') !== space.get('userNickname')) {
       channel.addMessage(channelMessage);
     }
+  },
+
+  /**
+   * Ask for a channel's attendance list (users currently joined)
+   *
+   * @param {Space} space
+   * @param {Channel} channel
+   * @public
+   */
+  observeChannel(space, channel) {
+    let observeMsg = buildActivityObject(space, {
+      '@type': 'observe',
+      target: channel.get('sockethubChannelId'),
+      object: {
+        '@type': 'attendance'
+      }
+    });
+
+    this.log('xmpp', 'asking for attendance list', observeMsg);
+    this.sockethub.socket.emit('message', observeMsg);
   },
 
   /**
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -161,7 +161,16 @@ export default Service.extend({
   },
 
   updateChannelUserList(message) {
-    const channel = this.getChannelById(message.actor['@id']);
+    let channel;
+    switch(message.context) {
+      case 'irc':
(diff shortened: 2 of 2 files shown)


skddc:
The multiple-connections issue has been fixed!

skddc:
@galfert See my last comment. This could actually work now...

galfert:
I shall give it a try again.

galfert:
While I can be connected to multiple IRC servers at the same time now, it still doesn't work with XMPP. Nothing seems to work when also connected to an IRC server. Although there seem to be no explicit error messages either.

When just connecting to a XMPP server alone, getting the channel user list works. My [PR to sockethub-platform-xmpp](https://github.com/sockethub/sockethub-platform-xmpp/pull/15), that was needed for this PR to work, has been merged. So I'm removing the [WIP]. I'll file a separate issue to sockethub for the multiple connections issue.

app/services/sockethub-xmpp.js
```diff
@@ -190,6 +190,26 @@ export default Ember.Service.extend({
...
+  /**
+   * Ask for a channel's attendance list (users currently joined)
+   *
+   * @param {Space} space
+   * @param {Channel} channel
+   * @public
+   */
+  observeChannel(space, channel) {
+    let observeMsg = buildActivityObject(space, {
+      '@type': 'observe',
```
skddc: This looks kind of weird. The term "observe" is from the IRC command, no? I don't think XMPP actually has "observe" for a channel. Also, just for normal humans "observe channel" would not suggest the action of "get attendance list" imo.

app/services/sockethub-xmpp.js
```diff
@@ -94,8 +94,8 @@ export default Ember.Service.extend({
   handleJoinCompleted(space, message) {
     const channelId = message.target['@id'].split('/')[0];
     const channel = space.get('channels').findBy('sockethubChannelId', channelId);
-    if (!isEmpty(channel)) {
-      channel.set('connected', true);
+    if (channel) {
+      this.observeChannel(space, channel);
```
skddc: I don't see where the channel is set to connected in this PR. It looks like the observa

app/services/sockethub-xmpp.js
galfert: This observes the channel for the type 'attendance' (see 3 lines below). I don't know if this is IRC specific, but it's how Sockethub is expecting it.

app/services/sockethub-xmpp.js
galfert: The channel gets set to connected when receiving the channel's user list here: https://github.com/67P/hyperchannel/blob/fix/127-xmpp_room_user_list/app/services/coms.js#L175

skddc: Hmm, I think it's a bit unexpected that setting the user list sets the channel to connected as well.

skddc: Aight.

galfert:
> While I can be connected to multiple IRC servers at the same time now, it still doesn't work with XMPP.

I just tried again with @silverbucket's new `feature/rework-connection-handling` branches and with those I was actually able to connect to both IRC and XMPP at the same time \o/

PR #132, opened 2017-06-28
Indicate connection status in channel header

### LINKED ISSUE #131: Improve channel connection indication for mobile
Extracted from https://github.com/67P/hyperchannel/issues/118#issuecomment-300185940

> We also need visual indication of channel connection status in the channel UI itself, esp. on mobile screens.

galfert:
Closes #131

Shows the title for a disconnected channel with lower opacity.

### DIFF
--- a/app/components/channel-container/template.hbs
+++ b/app/components/channel-container/template.hbs
@@ -1,6 +1,6 @@
 <main>
   <header id="channel-header" {{action "menu" "global" "toggle"}} onClick="">
-    <h2 id="channel-name">{{channel.name}}</h2>
+    <h2 id="channel-name" class="{{if channel.connected "connected" "disconnected"}}">{{channel.name}}</h2>
     <p id="channel-title">{{channel.formattedTopic}}</p>
     <nav>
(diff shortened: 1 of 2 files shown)


skddc:
I'm pretty sure a lower-opacity header title is completely non-understandable if you don't know what it means.

galfert:
I thought it's a quick enhancement that at least people who know it from desktop have an indicator on mobile.

Any ideas how to improve it?
Maybe some kind of icon next to the channel name? Or a red dot next to it while it's disconnected?

skddc:
I think the only proper way is to actually indicate sent status on the messages and in the input field. But that kind of requires someone with UI/UX chops to come up with something good. It's not an easy thing to communicate without words I think.

A red dot would definitely be better for now than title opacity imo. We can also add a hint that appears on tap to explain it then. We need some kind of UI lib for hover/tap hints anyway.

skddc:
Merging this, because it's better than what we have, and nobody pushed another proposal for a long time. Sorry for the delay!

PR #134, opened 2017-09-08
Fix swipe events in Chrome

### LINKED ISSUE #133: Swipe event not working in Chrome
Opening/closing the sidemenu via swipe doesn't work anymore.

galfert:
Fixes #133

Defining an empty swipe handler on the #channel-content causes the handler on the app container component to be triggered again. Seems to be a bug on elements that have "overflow-y: auto" set, which prevents the event to be triggered otherwise.


skddc:
Nice catch!

PR #135, opened 2017-09-22
Update Ember to version 2.15

galfert:
Noteworthy changes are improved performance and the replacement of PhantomJS with headless Chrome.

I also updated the `ember-gestures` addon and switched back to the original package instead of our own fork. The issues we created the fork for (build errors with newer Ember versions) have been fixed upstream as well.


skddc:
Works like a charm. And actually feels a little bit snappier!

PR #136, opened 2018-01-09
Fix overflow problems on mobile

fsmanuel:
The PR adds `overflow: auto` to `#channel` and `main` and `overflow-wrap: break-word` to `#channel` to make it work on mobile.

As a side effect we can remove the `.chat-message a` wrap rules.


skddc:
Nice!

PR #139, opened 2018-01-13
Use password inputs for password fields

galfert:
I noticed that the space config forms used normal inputs for the password fields.


skddc:
Anyone has an idea about that Chrome error on Travis?

skddc:
Can we just go back to good old phantomjs? The whole new Chrome setup is creating problems everywhere we use it. :/

galfert:
The tests on Travis are all green now.

app/components/space-setting-modal/template.hbs
```diff
@@ -21,14 +21,11 @@
...
   <p>
-    {{input value=space.server.password placeholder="password"}}
+    {{input type="password" value=space.server.password placeholder="password"}}
   </p>
   <p>
     {{input value=space.server.nickname required="required" placeholder="nickname"}}
   </p>
-  <p>
-    username:  {{space.server.username}}
-  </p>
```
skddc: This is removing the username?

galfert: No, this was outputting the username below the form from the actual username input field. Don't know why this was there in the first place, I guess it was for debugging purposes.

skddc: Ah, ok.

PR #150, opened 2018-06-13
Web Push Notifications for username mentions

skddc:
For the default Freenode space to work with this, it still needs a change on the dev server to be configured via Chef, but I'm sitting in a train currently.

In general, this subscribes to notifications via the API provided by [hubot-web-push-notifications](https://github.com/67P/hubot-web-push-notifications/), a Hubot script which I wrote a short while ago, and which will notify subscribed users of mentions, when their username is not currently online. (That means it will not work if you're online via a bouncer or sth.)

One general issue we'll have to solve is how a client can automatically discover Kosmos agents/bots/logs. I has the idea of setting up some kind of transparent directory service, which is easy to contribute to. This would also apply to finding and easily connecting to servers/networks in general, not just special services on them, like these push notifications. Maybe someone else had similar ideas?

refs #2

### DIFF
--- a/app/components/web-push-subscription/component.js
+++ b/app/components/web-push-subscription/component.js
@@ -1,0 +1,71 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import urlBase64ToUint8Array from 'hyperchannel/utils/url-base64-to-uint8-array';
+
+export default Component.extend({
+
+  botkaURL: null,
+  username: null,
+  subscribed: false,
+
+  webPushBaseURL: computed('botkaURL', function() {
+    return this.botkaURL + '/web-push';
+  }),
+
+  init () {
+    this._super(...arguments);
+
+    navigator.serviceWorker.ready.then(registration => {
+      return registration.pushManager.getSubscription();
+    }).then(subscription => {
+      if (subscription) { this.set('subscribed', true); }
+    });
+  },
+
+  actions: {
+
+    subscribe () {
+      navigator.serviceWorker.ready.then(async function(registration) {
+        // Get the server's public key
+        const response = await fetch(this.webPushBaseURL+'/vapid-public-key');
+        const vapidPublicKey = await response.text();
+        // Chrome doesn't accept the base64-encoded (string) vapidPublicKey yet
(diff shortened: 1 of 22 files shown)


app/components/web-push-subscription/component.js
```diff
@@ -0,0 +1,71 @@
...
+        const vapidPublicKey = await response.text();
+        // Chrome doesn't accept the base64-encoded (string) vapidPublicKey yet
+        // urlBase64ToUint8Array() is defined in /tools.js
+        const convertedVapidKey = urlBase64ToUint8Array(vapidPublicKey);
+        // Subscribe the user
+        return registration.pushManager.subscribe({
+          userVisibleOnly: true,
+          applicationServerKey: convertedVapidKey
+        });
+      }.bind(this)).then(subscription => {
```
galfert: Do fat arrow functions not work with `async`?

skddc: Oh, this might be a leftover from the code I copied. Might work with a fat arrow, too.

app/components/web-push-subscription/component.js
```diff
@@ -0,0 +1,71 @@
...
+
+  actions: {
+
+    subscribe () {
+      navigator.serviceWorker.ready.then(async function(registration) {
+        // Get the server's public key
+        const response = await fetch(this.webPushBaseURL+'/vapid-public-key');
+        const vapidPublicKey = await response.text();
+        // Chrome doesn't accept the base64-encoded (string) vapidPublicKey yet
+        // urlBase64ToUint8Array() is defined in /tools.js
```
galfert: This comment doesn't seem to be accurate.

config/space-presets.json
```diff
@@ -35,6 +36,20 @@
...
+    "name": "IRC localhost",
+    "description": "Local IRC server for development",
+    "website": "",
+    "protocol": "IRC",
+    "server": {
+      "hostname": "localhost",
+      "secure": false,
+      "port": 6667
+    }
+    },
```
galfert: This looks like a superfluous brace.

skddc:
I'll deploy it from this branch before merging. Thanks for having a look!

PR #151, opened 2018-08-13
Update npm dependencies

gregkare:
* [x] Update ember-cli-sass to 7.1.7 for Node 8 compatibility
* [x] Security updates

### DIFF
--- a/vendor/sh-assets-local.html
+++ b/vendor/sh-assets-local.html
@@ -1,3 +1,3 @@
-<script src="http://localhost:10550/activity-streams.js"></script>
+<script src="http://localhost:10550/activity-streams.min.js"></script>
 <script src="http://localhost:10550/socket.io.js"></script>
 <script src="http://localhost:10550/sockethub-client.js"></script>


skddc:
Isn't the second commit completing the second open task?

gregkare:
> Isn't the second commit completing the second open task?

Almost, ember-polyfill-for-tests and ember-service-worker depend on versions of iodash and minimatch that have vulnerabilities, but that will need an Ember update, so I think it should be done in a separate issue

This is ready to merge

skddc:
> but that will need an Ember update

Am I missing something, or is this PR not also updating Ember?

gregkare:
> Am I missing something, or is this PR not also updating Ember?

Never mind, I completely misunderstood. ember-polyfill-for-tests and ember-service-worker don't have newer versions, so this PR is doing all the updating we can do

skddc:
> ember-polyfill-for-tests and ember-service-worker don't have newer versions, so this PR is doing all the updating we can do now

1. Do we still need the polyfills?
2. Why can't we fork ember-service-worker, update the unsafe dependency, and use our own safe version in addition to creating a PR for them?

gregkare:
Good point about the ember-polyfill-for-tests, it was for an old version of PhantomJS (and we don't use PhantomJS), I removed it!

I checked, a PR has been merged in ember-service-worker (https://github.com/DockYard/ember-service-worker/issues/101) but no release so far

skddc:
Alright, great.

skddc:
I added `ember-cli-update` and tried to run it with codemods, but I wasn't able to get it to actually update Ember. @galfert Any idea how this works or what could be broken?

galfert:
I'm not sure, I haven't used it in quite a while.

Are you using it as a global executable via `ember-cli-update`? If I remember correctly, using it as a local ember addon via `ember update` never worked for me. E.g. even `ember update --list-codemods` didn't list anything.

Also, I think sometimes I had to explicitly state the versions with `--from` and `--to` to make it work.

skddc:
Could you please try it on your machine? As I said on IRC, I tried *everything* and I can't get anything to be updated.

galfert:
I updated to the latest stable Ember version (3.7.0) and also updated some other packages to fix most of the deprecation warnings.

There's still a few deprecation warnings from some packages that don't have a fix yet, but have open issues about it:

ember-browserify: https://github.com/ef4/ember-browserify/issues/130
ember-route-action-helper: https://github.com/DockYard/ember-route-action-helper/issues/86

There is also one linting error for a template left. First it was an error from the [no-invalid-interactive](https://github.com/ember-template-lint/ember-template-lint/blob/0a80191380e0074e8b8c3eee72171f46af137277/docs/rule/no-invalid-interactive.md) rule, which I fixed by adding the suggested `role="button"` to the element. But now it triggers the [no-nested-interactive](https://github.com/ember-template-lint/ember-template-lint/blob/0a80191380e0074e8b8c3eee72171f46af137277/docs/rule/no-nested-interactive.md) rule, because there are nested `a` elements.

Btw, I think the Sockethub error is an actual Sockethub issue with the current master. The file `activity-streams.js` that we [linked here](https://github.com/67P/hyperchannel/blob/387982df69d188e1a47d76eb6b13009f7d47d91a/vendor/sh-assets-local.html#L1) doesn't exist at all anymore. And the Sockethub examples at http://localhost:10550 (when running SH with `bin/sockethub --examples`) also have the same errors (e.g. "SockethubClient is undefined").
Instead of master I tried an old branch (`feature/rework-connection-handling`) which has been deleted on the remote already. With that one the tests run fine.

I hope @silverbucket has an idea about that.

galfert:
> Great! Care to share what command/setup you used exactly?

I just ran `ember-cli-update --to 3.7` and then `ember-cli-update --run-codemods`

PR #159, opened 2019-03-12
Fix duplicate archive message loading

### LINKED ISSUE #103: Duplicate messages when loading archive logs
When hitting the "Load previous messages" twice very fast, the archive logs get loaded twice leading to duplicate messages showing up.

galfert:
Fixes #103

Using ember-concurrency prevents running the same task more than once at the same time.

The [ember-concurrency addon](http://ember-concurrency.com) is pretty
nice. Should be useful for a lot of other asynchronous tasks as well.


skddc:
Fix works. :+1:

PR #160, opened 2019-03-19
Allow to switch channels via keyboard shortcuts

### LINKED ISSUE #148: Keyboard shortcut for fast-switching channels
This is something I can't live without in my desktop IRC client. It supports ctrl-pgup/down for going up and down the channel list. Not sure what the best combo for a web app would be, and also not sure what a good compromise would be for users of all of macOS, Windows, GNOME and KDE.

galfert:
Closes #148

Ctrl+Shift+Down goes to next channel
Ctrl+Shift+Up goes to previous channel

### DIFF
--- a/app/components/channel-nav/component.js
+++ b/app/components/channel-nav/component.js
@@ -39,7 +39,7 @@ export default Component.extend({
 
   transitionToRelativeChannel (relativePosition) {
     if (isPresent(this.activeChannel)) {
-      let channels = this.activeChannel.space.channels;
+      let channels = this.activeChannel.space.sortedChannels;
       let currentPosition = channels.indexOf(this.activeChannel);
 
       let edge = channels.length-1;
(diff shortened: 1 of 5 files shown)


galfert:
I borked the original PR during a refactoring, it wasn't using the sorted channel list anymore. So the switching was all over the place. That's fixed now.

app/components/channel-nav/component.js
```diff
@@ -1,7 +1,69 @@
...
+    if (isPresent(this.activeChannel)) {
+      let channels = this.activeChannel.space.sortedChannels;
+      let currentPosition = channels.indexOf(this.activeChannel);
+
+      let edge = channels.length-1;
+      let edgeOpposite = 0;
+
+      if (relativePosition === -1) {
+        // swap variables
+        [edge, edgeOpposite] = [edgeOpposite, edge];
```
fsmanuel: Nice!

galfert: ES6 FTW :)

PR #162, opened 2019-04-29
Add partial message rendering

### LINKED ISSUE #92: Better loading/display of heavy-traffic channels
So, now that we log some heavy-traffic channels, I noticed that they take very long to load on my phone, and the scrollback is just way too long to be useful by default.

I think we need to limit how many messages are loaded by default and have the previous link adapt to be either previous day or previous messages, or even have a second one to load within a day. Not sure.

You can currently reproduce this with <del>#mastodon</del> #indieweb on Freenode for example (towards the evening).

galfert:
Closes #92

This displays only the last 30 messages when opening a channel to improve the rendering speed. More messages are then displayed when scrolling up to the top of the page.

When the user starts scrolling, automatic scrolling to the bottom on new messages is disabled. It's re-enabled when the user scrolls to the bottom again or when switching channels.

### DIFF
--- a/app/components/channel-container/component.js
+++ b/app/components/channel-container/component.js
@@ -16,7 +16,7 @@ export default Component.extend({
   newMessage: '',
   channel: null,
   automaticScrollingEnabled: true,
-  paginationObserver: null,
+  partialRenderingObserver: null,
   scrollingObserver: null,
 
   coms: service(),
@@ -26,7 +26,7 @@ export default Component.extend({
 
     scheduleOnce('afterRender', this, function () {
       this.createScrollingObserver();
-      this.createPaginationObserver();
+      this.createPartialRenderingObserver();
 
       // We need to define an empty handler for swipe events on the
       // #channel-content element, so that the actual handler of the app container
@@ -57,7 +57,7 @@ export default Component.extend({
   },
 
   // loads new messages when the last message comes into view
-  createPaginationObserver () {
+  createPartialRenderingObserver () {
     const config = {
       root: this.element,
       rootMargin: '0px',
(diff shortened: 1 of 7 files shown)


skddc:
Same as for the linked issue: I don't think this is pagination, because there aren't any pages. I think the more apt term is "partial rendering" or something along those lines.

galfert:
I replaced the `onscroll` event handler with another intersection observer.

I also added a variable to control the position of the observed message, so additional messages are rendered earlier when scrolling up, improving the jumping.

Lastly I renamed the variables, properties and methods to remove any references to "pagination".

skddc:
Great! This works much better already.

One bug that I immediately saw is that when you open a channel with a lot of messages now, the scroll position is not the bottom by default anymore. You can try this by opening e.g. #indieweb on freenode today.

galfert:
I fixed the problem with the scroll position not being set to the bottom on heavy traffic channels.

I also found a better and much simpler solution for when to trigger rendering of additional messages. Instead of using the `observedMessageOffset` variable to configure which message to observe, it's now observing the upmost message again. But the IntersectionObserver is now configured with a different `rootMargin`. This config defines a margin around the observed element, that will trigger the intersecting already before the element itself comes into view.

Right now it's configured to be 25% of the height of the channel container.

galfert:
@skddc did you have a chance to look at my latest changes yet?

galfert:
@gregkare @silverbucket care to take a look as well?

gregkare:
I have tried it, I had an issue with the logs for the #mastodon channel on Freenode. After clicking on "Load previous messages" to get more messages a couple times it stopped loading more halfway through 5/17/2019

```
TypeError: this.partialRenderingObserver is null
```

However after reloading the app I can't reproduce the problem again and everything works

skddc:
Found a new round of bugs/issues (testing with the IndieWeb channel again):

1. The threshold is now so small, that the loading happens only when reaching the very end of the messages.
2. After scrolling up, and then clicking load-more, it jumped back down to the last message.
3. After loading the user list and/or title, the last message jumped slightly out of frame, so that the scoll position wasn't at the very bottom anymore. This one should likely be solved outside of this PR.

galfert:
Replaced by #167.

PR #166, opened 2019-05-28
Use partial rendering for large user lists

### LINKED ISSUE #163: Add partial rendering for user list
Similar to #92, opening a channel with a lot of users (e.g. #debian on Freenode) is slow, because it takes so long to render the list of users.

To improve this, we can display only a certain number of users at first, and then display more when scrolling to the bottom of the list. See #162 for the implementation using an `InterectionObserver` for the channel's messages.

galfert:
Closes #163

To improve the performance when switching to channels with a lot of users, only a subset of the users is rendered at first. More users will be rendered, when scrolling to the bottom of the list.


galfert:
When reviewers think this works ok, I will replace #162 with a new solution using the scrolling-observer component from this PR.

PR #167, opened 2019-06-05
Partial message rendering

### LINKED ISSUE #92: Better loading/display of heavy-traffic channels

galfert:
Closes #92

This displays only the last 30 messages when opening a channel to improve the rendering speed. More messages are then displayed when scrolling up to the top of the page.

This PR is a replacement for #162 and should fix all the remaining problems that PR still had.

The code is also much cleaner/simpler. It uses the scrolling-observer component introduced in #166 both for rendering more messages when scrolling up, as well as enabling/disabling automatic scrolling when scrolling down.

### DIFF
--- a/app/components/channel-container/component.js
+++ b/app/components/channel-container/component.js
@@ -1,16 +1,13 @@
 /* global Hammer */
-import $ from 'jquery';
-
 import Component from '@ember/component';
-import { observer } from '@ember/object';
 import { scheduleOnce } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency';
+import { computed, observer } from '@ember/object';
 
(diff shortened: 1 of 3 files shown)


skddc:
There's still a small issue with it suddenly jumping up a bit when scrolling through long histories. But it's such a huge improvement over what we have, that I'd say let's merge it and do shaping afterwards, in separate PRs.

Great job, @galfert!

group-five

PR #168, opened 2019-06-11
Show own messages in different windows.

### LINKED ISSUE #117: User's own messages don't show up in other tabs/windows
When having Hyperchannel open in multiple tabs or browser windows with the same user, the user's message only shows up in the tab the user wrote the message in.

That's because we currently discard the Sockethub messages for the user and only use the submitted message directly.

One idea was to just show the message in a lighter gray when sent, and then show it normally when it was received, indicating that others should have received the message as well now.

galfert:
Closes #117

This change makes sure that your own messages are also shown when having the app open in another tab, window or browser.

It also fixes a bug of a missing date headline when writing the first message of the day in a channel

### DIFF
--- a/app/controllers/space/base_channel.js
+++ b/app/controllers/space/base_channel.js
@@ -29,7 +29,7 @@ export default Controller.extend({
         message.get('content')
       );
 
-      this.get('model.messages').pushObject(message);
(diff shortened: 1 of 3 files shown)


skddc:
Seems to work as advertised. :+1:

PR #170, opened 2019-06-11
Update ember-route-action-helper

galfert:
This fixes a deprecation warning about using a private API.


galfert:
Replaced with #171

PR #171, opened 2019-06-11
Refactor actions to remove ember-route-action-helper dependency

galfert:
Instead of updating the package to fix some deprecation warnings (#170), I did some refactoring to completely remove the need for the `route-action` helper.


skddc:
Noice!

PR #173, opened 2019-06-12
Improve production builds and deployment

skddc:
Removes all custom build & deploy scripts, switching to 5apps Deploy's new build directory config instead. Builds production to `release/` when updating the version using `npm version`.


galfert:
LGTM

PR #176, opened 2019-07-03
Update ember-service-worker

galfert:
This fixes the warning 

> The following options have been renamed — please update your config: exports -> output.exports 

that is shown in the terminal when starting the app.


skddc:
Nice. That's it for the warnings then. :)

PR #177, opened 2019-07-04
Add username completion

### LINKED ISSUE #26: Name completion
When I enter the beginning of a name for a user in the current channel in the message field and hit tab, the name should be completed.

galfert:
Closes #26

Just write the first letter(s) of a username and hit tab.

When the name is completed at the beginning, a colon is added after the name.

This doesn't support cycling through names when there are multiple users with the same prefix yet. I'll create a separate issue for it.

### DIFF
--- a/app/components/message-input/component.js
+++ b/app/components/message-input/component.js
@@ -1,0 +1,68 @@
+import Component from '@ember/component';
+import { bindKeyboardShortcuts, unbindKeyboardShortcuts } from 'ember-keyboard-shortcuts';
+import { isPresent } from '@ember/utils';
+import { scheduleOnce } from '@ember/runloop';
+
+export default Component.extend({
+
+  tagName: 'section',
+  elementId: 'new-message',
+
+  message: '',
+  usernames: null,
+
+  keyboardShortcuts: Object.freeze({
+    'tab': 'completeUsername'
+  }),
+
+  didInsertElement () {
+    this._super(...arguments);
+    bindKeyboardShortcuts(this);
+  },
+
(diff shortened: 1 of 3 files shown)


app/components/message-input/component.js
```diff
@@ -0,0 +1,68 @@
...
+
+    completeUsername () {
+      let input = this.element.querySelector('#message-field');
+      let cursorPosition = input.selectionStart;
+      let textBeforeCursor = this.message.slice(0, cursorPosition);
+      let textAfterCursor = this.message.slice(cursorPosition);
+      let words = textBeforeCursor.split(' ');
+      let searchWord = words.pop();
+
+      if (isPresent(searchWord)) {
```
skddc: I think CodeClimate has a point about this function. Why not just return here, instead of wrapping all code in nested ifs?

app/components/message-input/component.js
```diff
@@ -0,0 +1,68 @@
...
+
+  actions: {
+
+    completeUsername () {
+      let input = this.element.querySelector('#message-field');
+      let cursorPosition = input.selectionStart;
+      let textBeforeCursor = this.message.slice(0, cursorPosition);
+      let textAfterCursor = this.message.slice(cursorPosition);
+      let words = textBeforeCursor.split(' ');
+      let searchWord = words.pop();
```
skddc: I don't think all of these are being re-assigned afterwards, so most of them should be `const`, no?

skddc:
This is absolutely fantastic!

fuck-yeah-funk

PR #179, opened 2019-07-04
Update dependencies

skddc:
A little update orgy to bring everything halfway up to date.


galfert:
Looks good to me 👍 

But Travis failed when trying to run the linting.

skddc:
> But Travis failed when trying to run the linting.

Fixed the command, but actually the linting is already part of `npm test`, so I also removed the extra commands from the Travis config.

PR #181, opened 2019-12-23
Make Hyperchannel work with latest Sockethub

galfert:
Latest Sockethub version 3.1.2 got rid of the prefix requirement for IDs (e.g. "irc://").

At the moment the client doesn't receive any attendance list for channels when asking for it.

@silverbucket: any idea why that is? In the Sockethub logs I found this:

```
sockethub:platform:irc:JrkL343y0xszp2Vx unprocessed irc message:> :hitchcock.freenode.net 366 hyphen #hypertest :End of /NAMES list. +45ms
```


skddc:
@silverbucket Did you see @galfert's question?

silverbucket:
Oops, missed this.

@galfert My guess is that it's because we decided to change the name from 'group' to 'role'?
https://github.com/sockethub/sockethub/releases/tag/v3.1.2

If you're logging them, you should see incoming messages that you might not be handling due to the missing property? (I mean at the browser level)

That console message is irrelevant.

galfert:
Nope, I don't think the renaming from `group` to `role` is the reason. Hyperchannel is logging all incoming messages and I don't see any message with `@type='observe'` and the object's `@type='attendance'` like with the old Sockethub version.

silverbucket:
What about type `update`? 
https://github.com/silverbucket/irc2as/commit/6f7864c40e9dc1d56cd78a5a92a839050d3cb684#diff-b3daea57c92d86d9a22d5f718b781875L126-R127

silverbucket:
@galfert that should have linked to the tests which have all been updated from various `observe` or `update` types to all `update`. I think @gregkare and I did this back in Sept. and I failed to mention it in the release notes.

silverbucket:
@galfert If I remember correctly it was because, adding roles made it more practical to send 1 AS object per user rather than previously sending an array of usernames. Overall this is shrinking the different types of objects associated with various room information.

galfert:
Ok, I can see the separate `update` messages. But when a client sends an `attendance` request, shouldn't it rather get a full list of members of a channel? I think that's what both IRC and XMPP servers are doing, no?

galfert:
@silverbucket now using IRC works again, except for the missing `update` message for the own user.

Also, other that what I said in the conference call, the `published` field didn't just change from integer to string, but it was actually an ISO 8601 date string before. I created an issue for it at https://github.com/sockethub/sockethub/issues/247. I changed Hyperchannel to work with both formats, but personally I prefer the ISO 8601 format, because it's human readable.

app/services/sockethub-irc.js
```diff
@@ -89,6 +89,30 @@ export default Service.extend({
...
+    }
+
+    let channel = space.get('channels').findBy('sockethubChannelId', message.target['@id']);
+
+    if (isEmpty(channel)) {
+      console.warn('No channel for presence update message found. Creating it.', message);
+      channel = this.coms.createChannel(space, message.target['displayName'], message.target['@id']);
+    }
+
+    channel.set('connected', true);
```
skddc: Why does a presence update set the channel connection status (hundreds of times if there are hundreds of users)?

galfert: Before it was set when receiving the user attendance list. ATM I don't think there is any other message that would indicate that the user successfully joined the channel.

galfert:
@silverbucket for channels with a lot of users (e.g. #debian), the separate messages have a really bad performance impact.

skddc:
> @silverbucket for channels with a lot of users (e.g. #debian), the separate messages have a really bad performance impact.

I was actually going to ask about that. Even just a simple list of handles with symbols for roles can get very large.

galfert:
I added a hotfix so that the own username is added to the channel whenever receiving an attendance message for any other user. I'll create an issue for Sockethub to add back the attendance lists with all users of a channel.

This PR is ready to go now.

galfert:
Closing in favour of #192.

PR #183, opened 2020-01-10
Replace icon font with new SVG icons

### LINKED ISSUE #154: Switch to SVG icons instead of icon font
When I wanted to add a new icon the other day, I found out that we still use some ancient custom icon font, created on Icomoon, for which I don't even have an account anymore.

SVG icons are generally easier to work with, can be styled with CSS, animated, and so on. I came up with an easy way to use them in Ember last year: placing the icon's SVG in a `app/templates/icons/my-icon.hbs` file, and then simply including them as a partial via `{{partial "my-icon"}}`. (See https://gitlab.com/skddc/inspektor for a real-world example.)

Any comments/opinions/links on beautiful icon sets would be useful for this one, of course.

skddc:
Switching to SVG icons and the https://feathericons.com set.

Screenshot from 2020-01-10 10-30-31

closes #154


skddc:
Another power cut prevented me from pushing a fix for the failings specs (linter issue). All good to go now!

Issue #184, opened 2020-01-10
Upgrade codebase to Ember Octane


skddc:
We can do this bit by bit, because Octane (introduced in Ember 3.15) is backward-compatible with classic Ember features. It'll just throw a ton of linter warnings, of course.

Here's a handy cheat sheet for all the changes: https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/

skddc:
Just a heads-up: I started the update orgy on my machine...

skddc:
Note: `channel-container` component should be dissolved and most functionality moved into the `base-channel` route/controller.

PR #187, opened 2020-01-12
Update Ember (and dependencies)

skddc:
This updates Ember and all dependencies to the latest stable versions. It also updates syntax in a lot of places, and already ports some parts of the code to Octane features.

The existing template modules, which still need upgrading, are added to the `pending` array of the template lint config file, so that we only see warnings for existing code. New files will throw errors when using old syntax.

refs #184


app/components/message-input/component.js
```diff
@@ -60,10 +60,8 @@ export default Component.extend({
 
       // set the cursor right behind the inserted username,
       // but we have to wait for the update of the input first
-      scheduleOnce('afterRender', this, function () {
-        input.focus();
-        input.setSelectionRange(newCursorPosition, newCursorPosition);
-      });
+      scheduleOnce('afterRender', this, input.focus);
+      scheduleOnce('afterRender', this, input.setSelectionRange(newCursorPosition, newCursorPosition));
```
galfert: This is calling the `setSelectionRange` method directly, instead of scheduling it to the `afterRender` queue.

app/components/scrolling-observer/component.js
```diff
@@ -46,10 +44,8 @@ export default Component.extend({
             this.onIntersect();
           }
           if (this.enabled && this.retriggeringEnabled) {
-            scheduleOnce('afterRender', this, function () {
-              observer.unobserve(entry.target);
-              observer.observe(entry.target);
-            });
+            scheduleOnce('afterRender', this, observer.unobserve(entry.target));
+            scheduleOnce('afterRender', this, observer.observe(entry.target));
```
galfert: Same as above, this is calling the 2 methods directly instead of queuing them.

PR #189, opened 2020-01-15
Octanify channel-nav, message-input

skddc:
Needed to fork ember-keyboard-shortcuts to add Octane support.

Unit tests are not generally a thing for Glimmer components, and I think ours should actually be rewritten as acceptance tests. But to make it work for now, I used the solution described in
https://timgthomas.com/2019/11/unit-testing-glimmer-components/

refs #184

### DIFF
--- a/app/components/channel-nav/template.hbs
+++ b/app/components/channel-nav/template.hbs
@@ -1,5 +1,5 @@
-<ul {{did-insert this.didInsertNode}} {{will-destroy this.willDestroyNode}}>
-  {{#each this.spaces as |space|}}
+<ul {{did-insert this.bindKeyboardShortcuts}}
+    {{will-destroy this.unbindKeyboardShortcuts}}>
   {{#each @spaces as |space|}}
     <li>
       <h2><LinkTo @route="space" @model={{space}}>{{space.name}}</LinkTo></h2>
--- a/app/components/channel-nav/template.hbs
+++ b/app/components/channel-nav/template.hbs
@@ -1,5 +1,6 @@
 <ul {{did-insert this.didInsertNode}} {{will-destroy this.willDestroyNode}}>
   {{#each this.spaces as |space|}}
+  {{#each @spaces as |space|}}
     <li>
       <h2><LinkTo @route="space" @model={{space}}>{{space.name}}</LinkTo></h2>
       <ul>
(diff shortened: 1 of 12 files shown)


app/components/channel-nav/template.hbs
```diff
@@ -1,20 +1,22 @@
...
-    <ul>
-      {{#each space.sortedChannels as |channel|}}
-        <li class="{{if channel.connected "connected" "disconnected"}} {{channel.unreadMessagesClass}} {{if channel.visible "active" ""}}">
-          {{#if channel.isUserChannel}}
-            {{#link-to "space.user-channel" space channel}}{{channel.name}}{{/link-to}}
-          {{else}}
-            {{#link-to "space.channel" space channel}}{{channel.name}}{{/link-to}}
-          {{/if}}
-          <a {{action "leaveChannel" space channel}} class="leave-channel" title="Leave {{channel.name}}">x</a>
+<ul {{did-insert this.didInsertNode}} {{will-destroy this.willDestroyNode}}>
```
galfert: I think this reads pretty non-descriptive. It should rather say what it actually does, like 
`{{did-insert this.addKeyboardShortcuts}}` and `{{will-destroy this.removeKeyboardShortcuts}}`.

app/components/message-input/template.hbs
```diff
@@ -1,7 +1,9 @@
-<form {{action onSend on="submit"}}>
-  {{channel-input-field value=message
-                        placeholder="Type a message and hit enter"
-                        name="chat-message"
-                        id="message-field"
-                        autocomplete="off"}}
-</form>
+<section id="new-message" {{did-insert this.didInsertNode}} {{will-destroy this.willDestroyNode}}>
```
galfert: Same as above. I would suggest to use more descriptive method names.

app/components/channel-nav/template.hbs
skddc: I tend to disagree, because the only good function name would be exactly the same as the function name of the keyboard shortcuts library IMO. And then it's confusing which is which.

app/components/message-input/template.hbs
skddc: Same as above. Is this something that you wouldn't merge without?

app/components/channel-nav/template.hbs
galfert: The shortcut library uses `bindKeyboardShortcuts`, my suggestion was `addKeyboardShortcuts`.

skddc: But "add" doesn't mean anything in this context. "Bind" is the correct term.

PR #191, opened 2020-01-15
Octanify remaining services

skddc:


skddc:
Please note: the CodeClimate issues in this PR aren't new. It's just that the functions count as new, from transforming the module to a class.

PR #192, opened 2020-01-19
Modifications for latest Sockethub

galfert:
Continuation of #181, but based on latest changes in Hyperchannel master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "ember-cli-sri": "^2.1.1",
     "ember-cli-template-lint": "^1.0.0",
     "ember-cli-uglify": "^3.0.0",
-    "ember-concurrency": "^0.9.0",
+    "ember-concurrency": "^1.1.5",
     "ember-elsewhere": "^1.1.0",
     "ember-export-application-global": "^2.0.1",
     "ember-fetch": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "preversion": "npm test",
     "version": "npm run build-prod && git add release/",
     "postversion": "git push origin master",
-    "deploy": "git push 5apps master"
+    "deploy": "git push 5apps master",
+    "sockethub": "DEBUG=sockethub* ./node_modules/sockethub/bin/sockethub"
   },
   "engines": {
     "node": "8.* || >= 10.*"
@@ -87,7 +88,8 @@
     "qunit-dom": "^0.9.2",
     "remotestorage-module-kosmos": "^1.1.0",
     "remotestoragejs": "^1.2.2",
-    "sass": "^1.24.4"
+    "sass": "^1.24.4",
+    "sockethub": "^3.1.4"
   },
   "dependencies": {
     "extend": "^3.0.2"
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "version": "npm run build-prod && git add release/",
     "postversion": "git push origin master",
     "deploy": "git push 5apps master",
-    "sockethub": "DEBUG=sockethub* ./node_modules/sockethub/bin/sockethub"
+    "sockethub": "DEBUG=sockethub* sockethub"
   },
   "engines": {
     "node": "8.* || >= 10.*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,14 +33,13 @@ install:
   - cd $HOME
   - git clone https://github.com/sockethub/sockethub.git
   - cd sockethub
-  - git checkout v3.0.1
-  - npm ci
+  - yarn install
   - cd $TRAVIS_BUILD_DIR
   - npm ci
 
 before_script:
   - cd $HOME/sockethub
-  - npm start 2> /dev/null &
+  - yarn start 2> /dev/null &
 
 branches:
   only:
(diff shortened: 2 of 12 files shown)


skddc:
Unfortunately, this has a bunch of breaking issues when I run it on my machine with both Freenode IRC and an XMPP account:

The IRC connection is closed shortly after it has been established and the rooms have been joined:

```
  sockethub:platform:irc:Ittffkbq70ggEohU unprocessed irc message:> :cherryh.freenode.net 366 hyper_rau #kosmos-random :End of /NAMES list. +0ms
  sockethub:resource-manager flagging for termination platform instance Ittffkbq70ggEohU (no registered sockets found) +45s
  sockethub:resource-manager sockets: 0 instances: 1 +15s
  sockethub:resource-manager terminating platform instance Ittffkbq70ggEohU (flagged for termination: no registered sockets found) +0ms
  sockethub:platform:irc:Ittffkbq70ggEohU cleanup() called +16s
  sockethub:platform:irc:Ittffkbq70ggEohU irc client 'close' event fired. +137ms
  sockethub:worker:yQbDprtvgxRqNv-_AAAB caught platform domain error: Error: connection to server closed.
  sockethub:worker:yQbDprtvgxRqNv-_AAAB     at __forceDisconnect (/home/basti/src/sockethub/sockethub/node_modules/sockethub-platform-irc/index.js:589:11)
```

Something similar (probably the same) happens with XMPP. The connection is closed pretty much immediately after it has been established. But for XMPP none of the channel content or names show up in Hyperchannel, while for IRC it disconnects, but the space and channels all still show as connected.

```
  sockethub:platform:xmpp:uQ8kpRUJcXA3nek2 close received for sebastian@5apps.com/hyperchannel +489ms                                                                        
  sockethub:worker:yQbDprtvgxRqNv-_AAAB caught platform domain error: Error: received close event for sebastian@5apps.com/hyperchannel
  sockethub:worker:yQbDprtvgxRqNv-_AAAB     at EventEmitter.close (/home/basti/src/sockethub/sockethub/node_modules/sockethub-platform-xmpp/index.js:569:15)
```

galfert:
I can't reproduce this and have no idea what's going on there. Maybe @silverbucket has an idea?Also would be good if he and @gregkare can give it a try to see if it works for them.

skddc:
@galfert I can easily reproduce this when starting from clean local storage. Just choose nickname for IRC, it connects and works. As soon as you add the 5apps XMPP space it breaks.

gregkare:
I'm going to give this one a shot

gregkare:
Yup, I can reproduce this, as soon as I add the 5apps XMPP space it disconnects my connection to Freenode

skddc:
I remember this being the case in the past at some point. IIRC it was before 3.0.

galfert:
I tested with the Kosmos XMPP server and everything worked fine.

Just tested with the 5apps server and got the same problem. Found the reason to be a wrong port number in the preset setting (5223 instead of 5222). Fixed the port number and now it works.

I also added a preset for the Kosmos XMPP server.

skddc:
Great, thanks.

But this means that there's both a pretty mean bug in Sockethub (it shouldn't disconnect other platforms just because one of them fails to connect), as well as missing error handling for connections in Hyperchannel. So we need GitHub issues for both of those.

skddc:
Tested again, and now I don't see disconnects anymore. However, XMPP incoming and outgoing MUC messages still don't work for me. I did see the incoming messages from the server in master before, but also only the ones it sends upon joining, not normal ones.

Not sure if that means something is still broken, or if that's simply because the XMPP support isn't there yet.

skddc:
@galfert Did you see my question?

> Not sure if that means something is still broken, or if that's simply because the XMPP support isn't there yet.

galfert:
I don't know, but I think it's a problem of the Sockethub XMPP implementation rather than a problem on the client.

I just was able to reproduce the problem, where I joined a channel but couldn't receive or send any messages. And the Sockethub logs didn't show anything, no errors but also no events for e.g. receiving messages.

After I restarted Sockethub, it worked again.

skddc:
What do we do here? I also don't get messages in `master` most of the time, so I think it's not a new issue.

When we merge this, we also have to deploy latest Sockethub in production btw.

skddc:
@galfert Can you update this with the new release as a dev dependency, and an npm script for running it, as you proposed elsewhere?

galfert:
Sure.

galfert:
> Can you update this with the new release as a dev dependency, and an npm script for running it, as you proposed elsewhere?

Done.

package.json
```diff
@@ -28,7 +28,8 @@
     "preversion": "npm test",
     "version": "npm run build-prod && git add release/",
     "postversion": "git push origin master",
-    "deploy": "git push 5apps master"
+    "deploy": "git push 5apps master",
+    "sockethub": "DEBUG=sockethub* ./node_modules/sockethub/bin/sockethub"
```
skddc: As sockethub defines an executable to be installed, and executables from dependencies can be used in npm scripts, this should just read:

```suggestion
    "sockethub": "DEBUG=sockethub* sockethub"
```

skddc:
@galfert Did you actually try this? It's missing all the dependencies that contain the client libraries for me.

galfert:
I only tested if the server started without errors. Haven't opened Hyperchannel.

skddc:
I opened #201 about the dependency issue and merged this one, so we can continue work here with the latest Sockethub.

PR #195, opened 2020-01-25
Slide out channel sidebar on small screens

skddc:
Only includes the basic layout and functionality for now, i.e. only the userlist slides out, and the code has to be refactored a bit as soon as we actually add more sidebar views (like e.g. list of shared items, mentions, channel settings, and whatnot). On desktop, everything stays the same.

hyperchannel-channel-sidebar

(Sorry about the jittering in the video gif. Not sure what happened there.)


skddc:
Just FYI: I have a follow-up PR ready on my machine, which makes the channel sidebar routable and adds support for wider sidebars with more content as well. Would be nice to merge this today, so I can do a PR against master with that.

skddc:
By the way, please note that when testing this on mobile, there's also #197 showing up, which is not a problem caused by the code in this PR.

PR #204, opened 2020-04-02
Improve channel list/nav for XMPP spaces

skddc:
* Show user address instead of space name
* Group channels by MUC domain
* Show short channel name instead of full JID. Add a hash so they look familiar (like IRC and Slack channels)

Screenshot from 2020-04-01 20-06-36

### DIFF
--- a/app/components/channel-nav/template.hbs
+++ b/app/components/channel-nav/template.hbs
@@ -2,7 +2,15 @@
     {{will-destroy this.unbindKeyboardShortcuts}}>
   {{#each @spaces as |space|}}
     <li>
-      <h2><LinkTo @route="space" @model={{space}}>{{space.name}}</LinkTo></h2>
+      <h2>
+        <LinkTo @route="space" @model={{space}}>
+          {{#if (eq space.protocol "XMPP")}}
+            {{space.server.username}}
+          {{else}}
+            {{space.name}}
+          {{/if}}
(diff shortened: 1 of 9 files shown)


app/components/channel-nav/template.hbs
```diff
@@ -1,23 +1,51 @@
...
-            {{/if}}
-            <a {{on "click" (fn @onLeaveChannel space channel)}} class="leave-channel" title="Leave {{channel.name}}">x</a>
+    {{#if (eq space.protocol "IRC")}}
+      <li>
+        <h2>
+          <LinkTo @route="space" @model={{space}}>{{space.name}}</LinkTo>
+        </h2>
+        <ul>
+          {{#each space.sortedChannels as |channel|}}
+            <li class="{{if channel.connected "connected" "disconnected"}} {{channel.unreadMessagesClass}} {{if channel.visible "active" ""}}">
```
fsmanuel, line 11: Maybe we can extract this `<li>` into a subcomponent `channel-nav/channel` as it seems to be the same for `XMPP`.

skddc: Thanks! I'll leave it for now, as it does not regard the actual improvements, and the changes also don't add much technical debt IMO. Refactors in separate PRs are obviously invited and encouraged. ;)

PR #202, opened 2020-03-31
Update Ember and octanify all the things

galfert:
Closes #184 

This updates Ember to the latest stable version (3.17) and updates the whole codebase to use the latest and greatest of Ember Octane (e.g. native JS classes, decorators, angle bracket components, ...)

PR #205, opened 2020-04-03
Replace Ember observers with render modifiers

galfert:
Fixes #202

After the Octane upgrade, all observers were broken. Replacing them with render modifiers fixed the issues and also removed the last remaining deprecation warnings.

The ScrollingObserver component doesn't use any Ember observers anymore, but still contains the word "observer". So I had to add an ESLint config line to ignore the "ember/no-observers" rule to the component.js file.


skddc:
thumbs-up-gruffy

PR #209, opened 2021-02-21
Fix automatic message input on username clicks

### LINKED ISSUE #208: Clicking usernames in chat messages throws exception
Should prefix the message draft with a username mention, but results in:

    Uncaught TypeError: this.onUsernameClick is not a function

Correctly defining it as `this.args.onUsernameClick` is not enough, because the actual function is trying to manipulate data that is handed in via args as well, which is not allowed in Octane.

raucao:
fixes #208


galfert (approved):
Change looks all good 👍 

Just the channel-container tests need the new arg as well now.

raucao:
@galfert Right, thanks. The suite was actually just unit tests for the `addUsernameMentionToMessage` function, so I moved them to the right place now, where they're not using `args`.

PR #210, opened 2021-03-27
Update Ember to latest LTS version

raucao:
Latest stable breaks the builds, and I wasn't able to find the cause.


galfert (approved):
Looks good but it's missing a Kredits label :)

raucao:
> but it's missing a Kredits label

That's because it's a draft PR. The linter is still failing, so I need to either change the rules, or update some things to more modern syntax.

galfert (approved):
Looks good and works 👍 
I like the new modifiers ❤️

PR #217, opened 2021-04-10
Update dependencies, Ember things

raucao:
Yet again, a little upgrade bonanza. Leaving a bunch of new Ember linting errors as warnings/todos for later upgrading.

I also left the autoprefixer and postcss stuff out of this, in favor of the Tailwind introduction PR. <strike>And I have begun an `ember-qunit` upgrade to `5.0.0` in a follow-up branch to this one, as it requires quite some changes in the unit tests.</strike>


raucao:
There's a failure left on CI, which I have no idea where it's coming from, as it works perfectly fine locally and I also haven't seen it on CI with another Ember app and rs.js beta:

```plain
TypeError: Cannot call a class as a function
    at _classCallCheck (webpack://__ember_auto_import__/./node_modules/remotestoragejs/release/remotestorage.js?:35:99)
    at new u (webpack://__ember_auto_import__/./node_modules/remotestoragejs/release/remotestorage.js?:1166:9)
    at h._createSuperInternal (webpack://__ember_auto_import__/./node_modules/remotestoragejs/release/remotestorage.js?:17:283)
    at new h (webpack://__ember_auto_import__/./node_modules/remotestoragejs/release/remotestorage.js?:6753:29)
    at R.featuresLoaded (webpack://__ember_auto_import__/./node_modules/remotestoragejs/release/remotestorage.js?:5790:140)
```

@galfert Maybe you have an idea?

raucao:
Downgraded rs.js for now, so this can be solved in a separate PR.

raucao:
Had another look at qunit and got it into this PR after all now.

galfert:
> @galfert Maybe you have an idea?

I think I might have seen this before and if I remember correctly it was because of the `browsers.push('ie 11')` in `config/targets.js`.

galfert:
Btw I found this about the git+ssh URLs in the package-lock.json: https://github.com/npm/cli/issues/2610#issuecomment-780761122

If I understand correctly, this happens for Git/Github URLs without a SHA in the package.json.

raucao:
> If I understand correctly, this happens for Git/Github URLs without a SHA in the package.json.

Good to know! Let's just keep the `postinstall` script in there for convenience until they fix the bug, so that nobody has to think about it when linking branches. And also, it makes it much clearer what is being linked, of course.

raucao:
> I think I might have seen this before and if I remember correctly it was because of the `browsers.push('ie 11')` in `config/targets.js`.

That was it! :pray: 

Not sure why it's the default to build IE11 stuff only in CI and prod anyway. So in the bin it goes...

raucao:
@galfert Could you have another look?

galfert (changes requested):
Left some comments. The rest looks good to me and testing didn't reveal any problems either.

galfert (approved):
👍

PR #218, opened 2021-04-13
Set up Tailwind CSS with PurgeCSS and JIT compilation

raucao:
Also removes Bourbon entirely, and switches Sass compilation from `ember-cli-sass` to PostCSS with plugins.

With this setup, we can now use Tailwind for new styles, as well as migrate existing ones to Tailwind classes and components over time.


package.json
```diff
@@ -42,14 +42,14 @@
...
   "devDependencies": {
     "@67p/ember-service-worker-hubot-web-push-notifications": "^1.0.1",
+    "@csstools/postcss-sass": "^4.0.0",
     "@ember/optional-features": "^2.0.0",
     "@ember/render-modifiers": "^1.0.2",
     "@ember/test-helpers": "^2.2.5",
     "@glimmer/component": "^1.0.1",
     "@glimmer/tracking": "^1.0.4",
-    "autoprefixer": "^7.2.6",
+    "autoprefixer": "^10.2.5",
```
galfert, line 51: Is this library still needed? It looks like all references to it have been removed from `ember-cli-build.js`.

raucao: You're right! Tailwind apparently comes with autoprefixer by default.

raucao:
Removed the obsolete package...

galfert (approved):
Changes look ok and I couldn't find any problems during testing.

PR #220, opened 2021-04-18
Fix user settings in localStorage

### LINKED ISSUE #219: User settings in localStorage not saved anymore
Looks like the `ember-local-storage` add-on doesn't work anymore. Thus, when loading the app from the index route, it won't redirect correctly, and when removing the default space entirely, then you'll end up with a blank page when loading the app from index later.

raucao:
Switch from `ember-local-storage` to `localForage` (and thus also `IndexedDB` by default).

Contains a generic function which allows to extend a `defaultValues` object in the service with more default data to be set for empty local storages.

Testing this manually is possible by just seeing if loading the app from the root URL redirects to the last channel you had open or not (should not do on master, but should do with this fix).

fixes #219


fsmanuel (commented):
Hey @raucao just a couple of questions. Nothing blocking except the test comments.

Impressive how much dependencies this one addon removes: `+117 −32,944` 😲 /cc @bumi

fsmanuel:
Totally confused. You didn't even remove `"ember-local-storage": "^1.7.2",` from the `package.json` and just added a new addon and still have that crazy diff in `package-lock.json`.
When we build the kredits wrapper I was discussing with @bumi if one should get more kredits if the PR removes more than it adds. As an insentive to produce better software... =) That would be a big one for you @raucao ...

app/services/local-data.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  async setDefaultValues () {
+    Object.keys(defaultValues).forEach(async storeName => {
+      const store = this.stores[storeName];
+
+      Object.keys(defaultValues[storeName]).forEach(async key => {
+        const value = await store.getItem(key);
+        const defaultValue = defaultValues[storeName][key];
+        if (isEmpty(value)) { await store.setItem(key, defaultValue); }
```
galfert, line 31: Be aware that `forEach` doesn't wait for any of the asynchronous functions before iterating over the next element. So the `await` for the `store.setItem()` call is superfluous.

raucao: Not sure what the transpiler does, but this made it actually work.

tests/unit/services/local-data-test.js
```diff
@@ -0,0 +1,22 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Service | local data', function(hooks) {
+  setupTest(hooks);
+
```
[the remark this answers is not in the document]
raucao: I added a hook, but it's not pretty. `localforage` is throwing an uncaught IndexedDB error on dropping the database for me...

app/services/local-data.js
```diff
@@ -0,0 +1,36 @@
...
+  constructor () {
+    super(...arguments);
+    this.stores = {
+      userSettings: localforage.createInstance({ name: `hyperchannel:user-settings` })
+    }
+    this.setDefaultValues();
+  }
+
+  async setDefaultValues () {
+    Object.keys(defaultValues).forEach(async storeName => {
```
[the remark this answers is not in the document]
raucao: I have refactored the entire function and also how it's being called. Thanks.

app/services/local-data.js
raucao: I have refactored the function to use `for ... of`. Thanks

app/routes/index.js
```diff
@@ -1,14 +1,17 @@
...
 
-  userSettings = localStorageFor('user-settings');
+  @service localData;
+  @alias('localData.stores.userSettings') userSettings;
 
-  redirect () {
-    let currentSpace = this.userSettings.currentSpace || config.defaultSpaceId;
-    let currentChannel = this.userSettings.currentChannel || 'kosmos';
+  async redirect () {
+    const currentSpace = await this.userSettings.getItem('currentSpace') || config.defaultSpaceId;
```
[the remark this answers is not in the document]
raucao: Not anymore. Removed.

raucao:
CodeClimate is throwing a fit, but I disagree with it about the complexity of that function.

tests/unit/services/local-data-test.js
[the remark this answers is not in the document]
fsmanuel: 👍

app/routes/application.js
```diff
@@ -4,11 +4,14 @@ import { action } from '@ember/object';
...
 export default class ApplicationRoute extends Route {
 
+  @service localData;
   @service logger;
   @service coms;
 
-  beforeModel () {
+  async beforeModel () {
     super.beforeModel(...arguments);
+    await this.localData.setDefaultValues();
```
fsmanuel, line 13: ❤️

fsmanuel (approved):
Nice and clean!

app/services/local-data.js
```diff
@@ -0,0 +1,39 @@
...
+    window.localforage = localforage;
+    this.stores = {
+      userSettings: localforage.createInstance({
+        name: 'hyperchannel', storeName: 'userSettings'
+      })
+    }
+  }
+
+  async setDefaultValues () {
+    for (const storeName of Object.keys(defaultValues)) {
```
fsmanuel, line 26: What you can try is assign `const defaultValuesKeys = Object.keys(defaultValues);` outside the class...

fsmanuel: To make CodeClimate happy

raucao: Didn't help. :/

PR #221, opened 2021-04-22
Fix getChannel() function

raucao:
The `observe` objects have the actor being the person and the target being the channel to observe. No idea how this worked for IRC before, because it was failing for IRC now, while XMPP [had the properties switched](https://github.com/sockethub/sockethub/pull/374) in its messages.


silverbucket (approved):

silverbucket:
LGTM!

PR #223, opened 2021-07-17
Update dependencies

raucao:
I have removed some obsolete things in the process, and also removed Sockethub as a direct dependency for now. I don't think it's worth the hassle for development at the moment.

Also tried upgrading Tailwind, but 2.1.1 is the latest tag that works without build issues for now.


galfert (approved):
LGTM

Issue #212, opened 2021-03-31
Switch from Travis CI to GitHub Actions


raucao:
Or migrate the repo to the Kosmos Gitea and use our Drone CI. Perhaps with automatic Git mirror pushes to GitHub.

@67P/core Opinions?

raucao:
Gitea just added push mirrors in the 1.15 release. :tada: 

https://blog.gitea.io/2021/08/gitea-1.15.0-is-released/#add-push-to-remote-mirror-repository-15157

PR #227, opened 2021-07-30
Refactoring spaces to accounts, channels

raucao:
Introducing some big changes to the data model and routes (mostly paired with @galfert on the code).

We have removed the concept of spaces entirely, and more importantly, the nesting of channels within spaces. Instead, we now use the more common concept of chat accounts, which channels only carry a reference to.

In (persistent/remote) storage, we have added channels as a new document/object type, so they are now persisted in a much more powerful way than simply an array of strings within a space.

XMPP accounts already work again as before. But we still have some things to do in other places:

* [x] Refactor spaces to accounts, persist channels separately
* [x] Refactor XMPP service for the new data model
* [x] Adapt UI to new models and routes
* [x] Refactor IRC service for the new data model
* [x] Rewrite account settings entirely (using the new add-account components for adding new accounts)
* [x] Adapt all tests where necessary

I think some of the remaining tasks should probably be done in separate PRs against this one, because this is already very big.


raucao:
I have migrated the old Travis CI repo to travis-ci.com and opened a PR against this branch to fix the builds.

Unfortunately, it looks like we'll run out of free credits rather quickly. Also, things are even buggier and more unstable than previously. I had pretty much the most abysmal experience over there just now.

On the plus side, this restores our CI for now, at least for a while. We could manually apply for open-source credits on Travis now, or we can resolve #212 instead.

raucao:
Last red check has been resolved. I finally dealt with CodeClimate GitHub auth again, in order to be able to change check issue statuses there. Also removed some duplication in the code.

super-green

PR #232, opened 2021-08-22
New account settings UI

raucao:
This is an MVP of new the account settings. For now, it only lists accounts, and allows to delete them. Editing is thus simply delete/add for now, until the settings UI is more sophisticated.

It removes all existing settings code for spaces, and replaces it with a new settings component, which is rendered entirely in a modal dialog. Also ports the existing modal CSS to Tailwind in the process.

Screenshot from 2021-08-22 19-46-32


galfert (approved):
Very nice 👍

PR #233, opened 2021-08-23
Adapt tests from spaces to accounts/channels

raucao:
All green again, but since Travis broke, we currently don't run CI on GitHub. We can add back CI in a separate PR then.

refs #227


galfert (approved):
Looking good and all green for me.

Issue #126, opened 2017-06-19
Initiate XMPP direct message channel


galfert:
Right now it's only possible to send direct messages, when the other user sends the first message.

skddc:
Note: render room DM channels under the actual room in the sidebar, so it's clear that they're linked.

PR #242, opened 2021-08-28
(Re)add direct messages for IRC and XMPP

### LINKED ISSUE #241: Fix/re-add direct messages
Disabled the initiation links in #227, but also not able to receive DMs anymore since that refactor.

Edit: actually, DMs still work fine for IRC, but receiving fails for XMPP.

raucao:
Also re-adds the ability to open URLs for channels that aren't currently joined.

For XMPP, the DM channel will be scoped to, and rendered under, the channel it belongs to, with an icon that indicates the scope. (We can improve channel titles/info etc. later on.)

Screenshot from 2021-08-28 13-42-40

closes #241, #126


app/routes/user_channel.js
```diff
@@ -2,9 +2,17 @@ import BaseChannel from 'hyperchannel/routes/base_channel';
 
 export default BaseChannel.extend({
 
-  // FIXME must select acccount automatically
-  createChannelOrUserChannel: function(account, channelName){
-    return this.coms.createUserChannel(account, channelName);
+  createChannelOrUserChannel: function(account, channelId) {
+    let channel;
+    switch(account.protocol) {
```
silverbucket, line 7: Is there a way this could remain unified across protocols? What's the factor that makes it impossible at the moment and is it addressable?

raucao: That's not possible, because with IRC the direct message channel is to the username/nickname, whereas with XMPP, the channel DM is scoped to the channel (while there can also be an actual DM directly to the user).

galfert (approved):
Changes look ok to me and everything worked as expected.

In fact, this makes Hyperchannel even better than Beagle, because they don't even support channel DMs :)

PR #243, opened 2021-08-29
Change channel info/settings icon

raucao:
Most people are not able to change channel settings, so it's going to be more of an info tab than a settings one.


galfert (approved):
👍

PR #244, opened 2021-08-29
Pending status for outgoing XMPP messages

raucao:
This marks outgoing messages as pending, and then confirms that they were sent when received back from the room. It also fixes the problem of one's own messages missing in the chat history that is sent by XMPP servers upon joining a channel.

<del>Note: will probably fail with DMs (#242), because we would need message carbons for those. Hence the WIP, until DMs are merged.</del> Fixed for DMs.

---

P.S.: I think we'll want to have a timeout for showing messages as pending in the UI later on. Currently, there's a bit of a flash from switching the text color after a split second. And we can also add an actual text marker or other indicator, after a few seconds of it not having confirmed.

### DIFF
--- a/app/controllers/base_channel.js
+++ b/app/controllers/base_channel.js
@@ -16,14 +16,20 @@ export default class BaseChannelController extends Controller {
 
   @alias('application.showChannelMenu') showChannelMenu;
 
-  createMessage (message, type) {
-    return new Message({
+  createMessage (content, type) {
+    const message = new Message({
       type: type,
       date: new Date(),
       // TODO  nickname per channel
       nickname: this.model.account.nickname,
-      content: message
+      content: content
     });
+
+    if (this.model.protocol === 'XMPP') {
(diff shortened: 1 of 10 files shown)


app/controllers/base_channel.js
```diff
@@ -16,14 +16,20 @@ export default class BaseChannelController extends Controller {
...
+    const message = new Message({
       type: type,
       date: new Date(),
       // TODO  nickname per channel
       nickname: this.model.account.nickname,
-      content: message
+      content: content
     });
+
+    if (this.model.protocol === 'XMPP') {
```
galfert, line 28: Why only for XMPP? Wouldn't it be the same process for IRC?

raucao: Nope. You don't receive your own messages back from a channel in IRC.

galfert (approved):
Nice one 👍

PR #246, opened 2021-08-31
Send and handle leave messages for XMPP MUC

### LINKED ISSUE #239: Send "leave" message for XMPP MUC
Available now in SH: https://github.com/sockethub/sockethub/issues/531

galfert:
Closes #239

During testing I noticed that Sockethub actually sends "online" presence updates for incoming "unavailable" stanzas. I've created an issue for that at https://github.com/sockethub/sockethub/issues/564.

This PR also deletes the creation of the ActivityStream object when leaving an IRC room, because I confirmed that it's not needed.

### DIFF
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -173,7 +173,7 @@ export default class ComsService extends Service {
   leaveChannel (channel) {
     switch (channel.protocol) {
       case 'XMPP':
-        // TODO implement
+        this.xmpp.leave(channel);
         break;
       case 'IRC':
         this.irc.leave(channel);
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -171,14 +171,8 @@ export default class ComsService extends Service {
   }
 
   leaveChannel (channel) {
-    switch (channel.protocol) {
-      case 'XMPP':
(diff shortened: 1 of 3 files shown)


app/services/coms.js
```diff
@@ -173,7 +173,7 @@ export default class ComsService extends Service {
   leaveChannel (channel) {
     switch (channel.protocol) {
       case 'XMPP':
-        // TODO implement
+        this.xmpp.leave(channel);
```
raucao, line 176: We can actually remove the protocol switch now, and do something like this:

```js
    this.getServiceForSockethubPlatform(channel.protocol)
        .leave(channel);
```

raucao (commented):
Nice! Had just one nit to pick.

app/services/coms.js
galfert: Good point 👍

raucao (approved):
check

galfert:
I've also added handling of the incoming "unavailable" presence updates that were fixed in https://github.com/sockethub/sockethub/pull/565. Now everything should be working dandy.

raucao (approved):
Works like a charm! :rocket:

PR #248, opened 2021-09-01
Fix redirects causing duplicate setup method calls

galfert:
Triggering a transition from the beforeModel method in the application route will cause the beforeModel method to be called again on the next transition. Moving the transition call to the redirect method fixes this.


raucao (approved):
Good catch! Fixed it. :tada:

PR #251, opened 2021-11-25
Replace custom modal implementation

galfert:
This replaces our custom modal implementation using liquid-fire and ember-elsewhere with [https://github.com/simplabs/ember-promise-modals](ember-promise-modals).

Liquid-fire is not maintained anymore for quite a while now. Also it was the last dependency of Hyperchannel that required jQuery.

The new modal library is much simpler to use and uses pure CSS animations.

During testing I noticed that the left menu was overlapping the settings modal on mobile screens. Changing some z-index styles fixed that issue as well.


raucao (approved):
That code looks much better than before! Let's add ALL THE MODALS soon, starting with the new channel join dialog!

I have tweaked the styles and animations a little bit. I think this looks better than the default.

galfert:
Fixed the merge conflict and checked out your changes. New animations look much better indeed

PR #254, opened 2021-11-29
UI for joining channels

raucao:
Adds a basic new channel-join dialog with an account switcher and different components based on the chat protocol of the account.

Left to do:

* [x] Fix modal focus trap
* [x] Fix failing integration test
* [x] Add integration tests for the new `join-channel-$protocol` components

### DIFF
--- a/app/components/join-channel/template.hbs
+++ b/app/components/join-channel/template.hbs
@@ -3,7 +3,7 @@
   <div class="flex-grow">
     <label for="account" class="hidden">Account</label>
     <select id="account" name="account"
-            disabled={{this.accountSelectionDisabled}} tabindex="0"
+            disabled={{this.accountSelectionDisabled}}
             {{on "change" this.selectAccount}}
             class="block ml-auto pl-3 pr-10 py-2 text-base border-gray-500 bg-gray-700
                    focus:outline-none focus:ring-indigo-500 focus:border-indigo-500
--- a/app/components/join-channel/template.hbs
+++ b/app/components/join-channel/template.hbs
@@ -5,7 +5,7 @@
     <select id="account" name="account"
             disabled={{this.accountSelectionDisabled}}
             onchange={{action (mut this.selectedAccountId) value="target.value"}}
-            class="block pl-3 pr-10 py-2 text-base border-gray-600 bg-gray-700
+            class="block pl-3 pr-10 py-2 text-base border-gray-500 bg-gray-700
                    focus:outline-none focus:ring-indigo-500 focus:border-indigo-500
                    sm:text-sm rounded-md mr-0 {{if this.accountSelectionDisabled 'text-gray-300'}}">
     {{#each this.coms.accounts as |account|}}
@@ -16,7 +16,7 @@
   </div>
 </div>
 
-<div class="w-full">
-  {{selectedAccount.id}}
-  {{yield}}
+<div class="w-full p-6">
+  {{component this.componentNameForProtocol
+              account=this.selectedAccount}}
 </div>
--- a/app/components/join-channel/template.hbs
+++ b/app/components/join-channel/template.hbs
@@ -4,17 +4,19 @@
     <label for="account" class="hidden">Account</label>
     <select id="account" name="account"
             disabled={{this.accountSelectionDisabled}}
+            onchange={{action (mut this.selectedAccountId) value="target.value"}}
             class="block pl-3 pr-10 py-2 text-base border-gray-600 bg-gray-700
                    focus:outline-none focus:ring-indigo-500 focus:border-indigo-500
                    sm:text-sm rounded-md mr-0 {{if this.accountSelectionDisabled 'text-gray-300'}}">
     {{#each this.coms.accounts as |account|}}
-      <option selected={{if (eq account.id this.selectedAccount.id) "selected"}}>{{account.id}}</option>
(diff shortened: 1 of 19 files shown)


tests/integration/components/join-channel/component-test.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  test('Select different account', async function(assert) {
+    this.coms = this.owner.lookup('service:coms');
+    this.set('coms.accounts', [ ircAccount, xmppAccount ]);
+    this.set('selectedAccountId', ircAccount.id);
+    await render(hbs`<JoinChannel />`);
+
+    await select('select#account', xmppAccount.id);
+    assert.equal(this.selectedAccountId, xmppAccount.id, 'updates the selected account ID property');
```
raucao, line 48: This works in the browser, but I can't get the test to succeed.

@galfert Maybe you have an idea?

raucao:
OK, we have working join dialogs now (and they're obviously already much nicer than our placeholder prompt).

There's an issue, at least in FF, with the ember modal library preventing tabbing between inputs. I checked the focus-trap lib they use, and I don't see why our other inputs shouldn't be tabbable, especially when specifically adding `tabindex` attributes.

Other than that, there's one failing integration test, which looks to me like it should be green. And there are no integration tests for the actual join components at the moment.

@galfert Maybe you could have a look at the first two issues with a fresh set of eyes.

tests/integration/components/join-channel/component-test.js
galfert: In integration tests you can't access the internal properties of the component. The `this` is the outer context of the component, e.g. the controller that uses the component.

Usually you would pass an action to a component and then test that it's called with the correct argument(s). Or you test that the DOM changes according to the interaction with the component.

I think in this case the appropriate test would be to also click the join button and then test that `coms.createChannel` is called with the correct arguments.

I'll have a stab at that.

galfert:
> There's an issue, at least in FF, with the ember modal library preventing tabbing between inputs. I checked the focus-trap lib they use, and I don't see why our other inputs shouldn't be tabbable, especially when specifically adding `tabindex` attributes.

That's some weird behaviour. I can only cycle backwards with shift+tab, but not forwards. And only between the select and the input, but not the submit button. When I remove the `tabindex` attributes, I can also tab to the button, but still only backwards.

raucao:
> That's some weird behaviour. I can only cycle backwards with shift+tab, but not forwards. And only between the select and the input, but not the submit button. When I remove the `tabindex` attributes, I can also tab to the button, but still only backwards.

Yup, same here. Very unusual. The library is basically destroying perfectly fine browser behavior, by trying to trap focus in the modal window.

app/components/join-channel/template.hbs
```diff
@@ -4,7 +4,7 @@
     <label for="account" class="hidden">Account</label>
     <select id="account" name="account"
             disabled={{this.accountSelectionDisabled}} tabindex="0"
-            onchange={{action (mut this.selectedAccountId) value="target.value"}}
+            {{on "change" this.selectAccount}}
```
raucao, line 7: Ah, nice!

galfert:
There is some discussion about making it configurable to disable focus trap at https://github.com/simplabs/ember-promise-modals/issues/435.

One option would be to fork the library and use the patch from https://github.com/simplabs/ember-promise-modals/issues/435#issuecomment-961771874 until it's supported officially.

raucao:
So, I forked the library, applied the patch, and removed all `tabindex` attributes. It looks very simple, and the code is executed as expected. But the result is even weirder now: 

* You can absolutely not tab forward, but you can tab backward from the second input field, and all the way through the rest of the DOM outside of the modal now. But never forward.
* Without even touching anything about the rest of the code, but clicking outside of the modal as well as ESC do not close the modal anymore. No exception in the console.

comeon

galfert:
That was an interesting one :)

I added the ember-promise-modal to another project to see how it behaves there. Turns out that even with the focus-trap activated all was working fine and as expected.

So there had to be something different in Hyperchannel itself. After some digging I found the tab keyboard listener for the message field that we use for the username completion.

After changing it to only activate when the message field is focused, all is working fine and dandy now.

I reverted the ember-promise-modals dependency back to the official release and also introduced a `{{autofocus}}` modifier  to focus on the channel name field when opening the dialog.

raucao:
Good catch!

And we can likely make use of the focus trap with the chat input actually, as we will have several more inputs grouped in that component, for attachments and such.

package.json
```diff
@@ -90,6 +90,7 @@
     "tailwindcss": "^2.2.4"
   },
   "dependencies": {
+    "ember-autofocus-modifier": "^3.0.0",
```
raucao, line 93: @galfert This was probably an accident, and should be a dev dependency, right?

galfert: Oh, that's weird. I installed it with `ember install` and that usually adds it as a dev dependency. I'll change it.

raucao:
> EDIT: I'll tackle the integration test for the `join-channel-irc` component next.

Maybe do it as a new PR against this one, because the kredits medium label is starting to become a bit small, but I don't feel like it's going to be a large overall...

galfert:
> Maybe do it as a new PR against this one, because the kredits medium label is starting to become a bit small, but I don't feel like it's going to be a large overall...

ok

PR #255, opened 2021-12-03
Integration tests for join-channel-* components

galfert:
This also changes the comment style of the commented code in the `join-channel-xmpp` component to fix the linter errors about HTML comments.

After merging #254 should be all green.


raucao (approved):
Great! :+1:

PR #256, opened 2021-12-03
Update Ember to 3.24.6

galfert:
Update Ember to latest LTS, fix deprecations and linter warnings.


raucao:
Ooof, I do vehemently disagree with the prettier formatting changes to be honest. Not sure what to say other than that it looks horrible to me and I only see drawbacks, but no benefits for these exact changes.

app/components/add-chat-account-xmpp/component.js
```diff
@@ -17,46 +16,57 @@ export default class AddChatAccountXmppComponent extends Component {
...
 
       const account = await this.addAccount();
       this.addDefaultChannels(account);
       this.finishedSetup = true;
 
-      const firstChannel = this.coms.channels.filterBy('account', account).firstObject;
+      const firstChannel = this.coms.channels.filterBy(
+        'account',
+        account
+      ).firstObject;
```
raucao, line 64: This is so much less readable than the original, it's not even funny. How is this prettier than the original, very concise one line?

raucao (changes requested):
As there's no reasoning documented in the PR description for introducing the Prettier changes, I cannot really discuss them properly. I find almost all of them to be terrible for readability, and less elegant. The ones that add lots of new lines for arguments also make the functions longer than CodeClimate is happy with at the moment, thus making a different linter fail by making Prettier happy.

Happy to hear why these changes are necessary or wanted. Let's discuss.

galfert:
Prettier comes by default with Ember now. The related RFC with the reasoning can be found at https://emberjs.github.io/rfcs/0628-prettier.html.

I have to agree that I don't like some of the changes in this PR either. They were from automated fixes by the linter. I'll see if some more manual changes can make both the linters as well as the human eye happy :)

raucao:
OK, so it's added to the default blueprint, but the RFC does not actually present reasoning for having to have it in every Ember app, other than the normal reasons for choosing a solution like Prettier.

So, personally, I'd rather not add it at all than having to fight it all the time. The point of something like Prettier is that you submit 100% to its code style, so that your tools can automatically format everything.

galfert:
Replaced by #258.

PR #259, opened 2021-12-07
Adapt to new AS2.0 names

### LINKED ISSUE #257: Adapt to new AS2.0 names
Sockethub switched over to AS2.0 names in sockethub/sockethub#574.

galfert:
Closes #257

This PR makes the following renames to be compatible with latest SH master:

@id -> id
@type -> type
displayName -> name

At the moment there is an error when querying the room attendance list though. I reported it in sockethub/sockethub#638


galfert:
Found and fixed a bug with an inconsistent actor ID. Now, when using the fix from sockethub/sockethub#641 XMPP finally works properly, and there is no reload necessary anymore after adding a new account.

There is still an issue with IRC though, because the messages sent by Sockethub still contain the old AS names. I created a new issue for that at sockethub/sockethub#646.

galfert:
With sockethub/sockethub#648 everything is working now. 

This PR is ready for review.

raucao (approved):
Hooray for the XMPP login flow working properly now!

PR #261, opened 2021-12-24
Add incoming message corrections

### LINKED ISSUE #245: XEP-0308: Last Message Correction
[XEP-0308: Last Message Correction](https://xmpp.org/extensions/xep-0308.html)

raucao:
Works in combination with sockethub/sockethub#649 (and will simply show additional messages when used with Sockethub versions that do not support it).

Adds the stanza ID as a new, separate property for message objects (needed for MAM), while the Sockethub change switches the ActivityStream object's "id" property to the XMPP `<message>` element's "id" attribute.

Left to do:

* [x] Only allow correcting the last message
* [x] Correction must come from the sender of the original message
* [x] Indicate in the UI that a message has been edited/replaced

refs #245


galfert (approved):
Changes look good and confirmed it to be working 👍

PR #263, opened 2022-01-03
Add outgoing message corrections

raucao:
Works in combination with sockethub/sockethub#653. (When used with Sockethub versions that don't support outgoing corrections it will just send a new message.)


tests/unit/services/sockethub-xmpp-test.js
```diff
@@ -142,4 +143,48 @@ module('Unit | Service | sockethub xmpp', function(hooks) {
...
+
+    xmpp.transferMessage(channel, message);
+
+    assert.ok(socketEmitSpy.calledOnce, 'emits a sockethub job message');
+
+    const jobMessage = socketEmitSpy.getCall(0).args[1];
+    assert.equal(jobMessage.context, 'xmpp', 'job context is correct');
+    assert.equal(jobMessage.type, 'send', 'job type is correct');
+    assert.equal(jobMessage.object.type, 'message', 'job object type is correct');
+    assert.equal(jobMessage.object.content, 'Only 4 days until 2022!', 'job object content is correct');
```
galfert, line 166: already outdated ;)

galfert (approved):
Works as expected.

But was able to produce some odd behavior:

1. Write a message.
2. Press the edit button for that message.
3. Don't edit or cancel yet, but send a new message from the normal message field.
4. Edit and send your first message.

Result: Hyperchannel shows the original (unedited) first message, but other clients show the edited message.

I don't think it's important to prevent that behaviour, but I'm wondering why Hyperchannel isn't showing the message correction.

tests/unit/services/sockethub-xmpp-test.js
raucao: Still fits the commit date. :)

raucao:
> I don't think it's important to prevent that behaviour, but I'm wondering why Hyperchannel isn't showing the message correction.

Yeah, that's as edge as a case can get. I agree that it's not important to account for that. But thanks for testing so thoroughly!

PR #267, opened 2022-01-09
Grouped chat messages

raucao:
Doesn't render every line of chat with all metadata, when the message is from the same user and in reasonably quick succession to the previous message.

I used 120 seconds as a cutoff for groupings. Not sure if that's too much already.

(The UI will be much nicer than this, and it will be even easier for the brain to parse messages belonging together, when we add avatars (with placeholders) soon.)


raucao:
I opened #268 against this one, which extends it with avatar placeholders.

galfert (approved):
👍

PR #268, opened 2022-01-09
Add user avatar placeholders

raucao:
Adds avatar placeholder icons with the first letter of the message sender's nickname, and coloring them according to [XEP-0392](https://xmpp.org/extensions/xep-0392.html).


galfert (approved):
Looks good to me.

PR #270, opened 2022-01-16
Adapt code to new Sockethub Client API

silverbucket:
I started testing the sockethub refactor (https://github.com/sockethub/sockethub/pull/596) with hyperchannel, and managed to fix a few things, however couldn't figure out how the join process was supposed to work after the connect (the connect works, just nothing is joined afterward).

Would be up to either pair with someone on finishing this, or if you just want to use this as an example branch and make a new PR that's fine with me too. The most important part to get started is to make sure you remove the <script> tags which include activity-streams.js as that's now a part of the sockethub-client.js

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ install:
   - cd $HOME
   - git clone https://github.com/sockethub/sockethub.git
   - cd sockethub
+  - git checkout refactor-sockethub-client
   - yarn install
   - yarn run build
   - cd $TRAVIS_BUILD_DIR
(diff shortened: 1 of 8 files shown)


raucao:
Nice!

Regarding joins, almost at the top of the changes in this branch, the code that handles completed joins is being removed entirely.

silverbucket:
> Regarding joins, almost at the top of the changes in this branch, the code that handles completed joins is being removed entirely.

Yeah I saw that, however that's still for completed joins, I didn't see where joins were being requested after a connect.

PR #274, opened 2022-02-06
Use new Sockethub IRC connect flow, show connect errors

### LINKED ISSUE #194: Test server config/connection before creating/saving a new account
Validate that the config actually results in a successful connection before adding the space to both RS and the UI.

* [x] XMPP
* [x] IRC

refs #192

raucao:
Implements and relies on https://github.com/sockethub/sockethub/pull/671

closes #194

### DIFF
--- a/app/components/add-chat-account-irc/component.js
+++ b/app/components/add-chat-account-irc/component.js
@@ -26,40 +26,6 @@ export default class AddChatAccountIrcComponent extends Component {
     return this.nickname;
   }
 
-  // TODO Implement once available
-  // async handleConnectFailure (message) {
-  //   console.debug('handleConnectFailure called with', message);
-  // }
-
-  // TODO Implement once available
-  // async handleConnectMessage (message) {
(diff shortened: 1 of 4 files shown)


raucao:
@silverbucket For some reason, the yarn build is failing with Sockethub master now:

    The command "yarn run build" failed and exited with 127 during .

raucao:
I have found the problem. For some reason, the standard `yarn install` in sockethub's root `package.json` does not run the lerna bootstrap/install anymore. I changed it to the new npm script, but I think this is less clean and intuitive than a standard `yarn install && yarn start`.

silverbucket:
@raucao yeah, unfortunately this is a limitation of `lerna` when there is a package.json in the root of the repository (in additon to each packages package.json). I created an issue here explaining it https://github.com/lerna/lerna/issues/3070, in the meantime we're going to have to use `yarn deps` as the one-step command to install all dependencies.

silverbucket (approved):
LGTM

PR #277, opened 2022-02-08
Add SASL authentication for IRC

### LINKED ISSUE #240: Authenticate via SASL (instead of NickServ)
Modern IRC networks support SASL authentication upon connecting, so you don't have to message NickServ after connecting. In fact, some of them *require* SASL when coming from certain IP ranges. I assume our production Sockethub instance would land on those lists rather quickly.

Depends on https://github.com/sockethub/sockethub/issues/543. Could supercede #32.

raucao:
For now, we just assume that all authenticated connections use SASL.

I was able to use this to successfully connect to Libera from my VPN, which they require SASL for. :tada: 

Tested with and depends on https://github.com/sockethub/sockethub/pull/673. Also required adding the `sasl` property to the RS module's schema (already published as 2.2.0).

closes #240

### DIFF
--- a/app/components/add-chat-account-irc/component.js
+++ b/app/components/add-chat-account-irc/component.js
@@ -14,29 +14,35 @@ export default class AddChatAccountIrcComponent extends Component {
   @tracked hostname = 'irc.libera.chat';
   @tracked port = '6697';
   @tracked nickname;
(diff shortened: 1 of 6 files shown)


silverbucket (approved):
Tried and seems to be working great, LGTM!

PR #282, opened 2022-02-12
Handle unified schema

silverbucket:
This ended up being tiny, not sure if I hit all the handling but so far I've just had to really update to ignore service messages (these are messages from the IRC server that are not from any one room, and the UI doesn't have an interface for handling). Reference https://github.com/sockethub/sockethub/pull/684

### DIFF
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -424,12 +424,7 @@ export default class ComsService extends Service {
       case 'update':
         switch (message.object.type) {
           case 'topic':
-            if (message.actor['type'] === 'service') {
-              // TODO (could also create a special service room)
-              // this.handleServiceAnnouncement()
-            } else {
-              this.updateChannelTopic(message);
-            }
+            this.updateChannelTopic(message);
             break;
           case 'address':
             this.updateUsername(message);
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -395,6 +395,11 @@ export default class ComsService extends Service {
   handleSockethubMessage (message) {
     this.log(`${message.context}_message`, 'SH message', message);
 
+    if (message.actor.type === 'service') {
+      this.log(`${message.context}_message`, 'skipping service message');
+      return;
+    }
+
     switch (message.type) {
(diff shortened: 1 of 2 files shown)


app/services/sockethub-irc.js
```diff
@@ -124,7 +124,7 @@ export default class SockethubIrcService extends Service {
     channel.addUser(account.nickname);
     channel.connected = true;
 
-    channel.addUser(message.actor.name);
+    channel.addUser(message.actor.name || message.actor.id);
```
raucao, line 127: The ID would include the service domain, so that should not be shown in a channel. What's the reason for this fallback? Wouldn't it be a bug if the nickname weren't available as `name`?

app/services/coms.js
```diff
@@ -395,6 +395,11 @@ export default class ComsService extends Service {
   handleSockethubMessage (message) {
     this.log(`${message.context}_message`, 'SH message', message);
 
+    if (message.actor.type === 'service') {
+      this.log(`${message.context}_message`, 'skipping service message');
+      return;
+    }
```
raucao, line 401: Service update messages are already handled in the `update` case below. Are there any new message types from the service that we should handle as well?

silverbucket: Well as the schema is defined now, it's not required. The only thing required is `id` and `type`.

silverbucket: Yeah I think this section should be improved, my skipping isn't a good solution. The handling further down of the `service` type is only in the case of `update`s, however sometimes there'll be a `send` from the `service`. Hyperchannel was erroring trying to display it like a person writing to a room. 

So in this case, it's a message from the service, with no room. We should either display those some other way or just ignore them or write them to the console. The `send` case should verify first that the target is a `room` before calling `addMessageToChannel`

app/services/sockethub-irc.js
raucao: OK, thanks. I changed it to extract the nickname from the ID when applicable.

PR #288, opened 2022-03-31
Fix CI builds

raucao:


raucao:
All working now. Aside from having to adapt some things, I found an actual bug in the tests, which was revealed by the new/improved Sockethub validations.

I cleaned up the history in this branch just now. I think it's good to go now.

PR #301, opened 2025-10-14
Only set IRC password in credentials if defined

silverbucket:


app/services/sockethub-irc.js
```diff
@@ -78,7 +78,7 @@ export default class SockethubIrcService extends Service {
         sasl: account.server.sasl,
         nick: account.nickname,
         // username: account.username,
-        password: account.password
+        ...(account.password && { password: account.password }),
```
raucao, line 81: If the property is `undefined`, then the password there would be set to `undefined`, i.e. not set. Or am I missing something?

silverbucket: This change ensures that if the password is not set, the property doesn't exist.

raucao:
This fix was included in #303

PR #308, opened 2026-06-01
Query and display MUC room info

raucao:
Query MUC room info when joining XMPP rooms, then add the results to the `BaseChannel` instance.

The settings pane is now protocol-specific, and the XMPP one will show basic room info, configs, and raw feature keys once loaded:

Screenshot From 2026-06-01 23-41-32

The "supported features" will have to be transformed into proper UI in a future PR.

Also, the channel display name is now using the configured name in the channel header as well as the titles for the channel leave links.

---

:warning: Depends on https://github.com/sockethub/sockethub/pull/916
refs #292

### DIFF
--- a/app/models/base_channel.js
+++ b/app/models/base_channel.js
@@ -18,6 +18,7 @@ export default class BaseChannel {
   @tracked unreadMessages = false;
   @tracked unreadMentions = false;
   @tracked visible = false; // Current/active channel
+  @tracked roomInfoLoaded = false;
   @tracked description = null;
   @tracked roomFeatures = new TrackedArray([]);
   @tracked roomIdentities = new TrackedArray([]);
--- a/app/models/base_channel.js
+++ b/app/models/base_channel.js
@@ -18,6 +18,12 @@ export default class BaseChannel {
   @tracked unreadMessages = false;
   @tracked unreadMentions = false;
   @tracked visible = false; // Current/active channel
+  @tracked description = null;
+  @tracked roomFeatures = new TrackedArray([]);
+  @tracked roomIdentities = new TrackedArray([]);
+  @tracked roomInfoData = null;
+  @tracked roomConfigData = null;
+  @tracked roomCustomData = null;
 
   constructor (props) {
     Object.assign(this, props);
--- a/app/services/coms.js
+++ b/app/services/coms.js
@@ -212,6 +212,42 @@ export default class ComsService extends Service {
     }
   }
 
+  updateChannelRoomInfo (message) {
+    this.log('xmpp', 'Room info update:', message.actor?.id, message.object);
+    const channel = this.getChannel(message.actor.id);
+
+    if (channel) {
+      channel.connected = true;
+      if (message.error) {
+        console.warn('Error querying room info:', message.error);
+        return;
+      }
+      if (message.actor && message.actor.name) {
+        channel.displayName = message.actor.name;
+      }
+      if (message.object) {
+        if (Array.isArray(message.object.features)) {
+          channel.roomFeatures = message.object.features;
+        }
+        if (Array.isArray(message.object.identities)) {
+          channel.roomIdentities = message.object.identities;
+        }
+        if (message.object.roominfo) {
+          channel.roomInfoData = message.object.roominfo;
+          if (message.object.roominfo.description && message.object.roominfo.description.value) {
+            channel.description = message.object.roominfo.description.value;
+          }
+        }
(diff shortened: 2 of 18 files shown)


app/services/sockethub.js
```diff
@@ -11,6 +11,17 @@ export default class SockethubService extends Service {
...
+
+    // Downgrade client-side validation errors to warnings to facilitate live testing of unreleased/custom schemas
+    const originalValidate = client.validateActivity.bind(client);
+    client.validateActivity = function (activity) {
+      const error = originalValidate(activity);
+      if (error) {
+        console.warn('[sockethub-client] Bypassing schema validation error:', error, activity);
+      }
+      return '';
+    };
```
silverbucket, line 23: This should be removed, I guess it was added before sockethub had room-info defined in schemas.

app/services/sockethub-xmpp.js
```diff
@@ -248,6 +249,34 @@ export default class SockethubXmppService extends Service {
...
+      },
+      object: {
+        type: 'room-info'
+      }
+    });
+
+    this.log('xmpp', 'asking for room info', msg);
+    this.sockethubClient.socket.emit('message', msg, (response) => {
+      if (response && response.error) {
+        console.error('[xmpp] query (room-info) rejected by Sockethub:', response.error);
```
silverbucket: are we intentionally not using the `this.log` here?

app/services/sockethub-xmpp.js
```diff
@@ -248,6 +249,34 @@ export default class SockethubXmppService extends Service {
...
+      } else {
+        console.log('[xmpp] query (room-info) accepted by Sockethub:', response);
```
silverbucket: are we intentionally not using the `this.log` here?

app/models/base_channel.js
```diff
@@ -18,6 +18,13 @@ export default class BaseChannel {
   @tracked unreadMessages = false;
   @tracked unreadMentions = false;
   @tracked visible = false; // Current/active channel
+  @tracked roomInfoLoaded = false;
+  @tracked description = null;
+  @tracked roomFeatures = new TrackedArray([]);
+  @tracked roomIdentities = new TrackedArray([]);
+  @tracked roomInfoData = null;
+  @tracked roomConfigData = null;
+  @tracked roomCustomData = null;
```
silverbucket, line 27: Just want to confirm that these should all be in the base channel and not protocol specific for xmpp? (will IRC use them?)

app/services/coms.js
```diff
@@ -212,6 +212,43 @@ export default class ComsService extends Service {
     }
   }
 
+  updateChannelRoomInfo (message) {
+    this.log('xmpp', 'Room info update:', message.actor?.id, message.object);
+    const channel = this.getChannel(message.actor.id);
+
+    if (channel) {
+      channel.connected = true;
```
silverbucket: Should this be here at all? (it's already handled in `updateChannelUserList`) and definitely not before the error handling.

If room-info errors and attendance also hasn't responded yet, the channel would be marked connected with no room data.

silverbucket (changes requested):

app/services/sockethub.js
raucao: No, I actually had to add it when switching to the query verb, because it wouldn't allow me to send the unknown "room-info" type object.

app/services/sockethub-xmpp.js
raucao: I changed it to `console.warn`, which we use for unexpected problems throughout the rest of the class.

app/services/sockethub-xmpp.js
raucao: Removed this one entirely, because it was just for debugging.

app/models/base_channel.js
raucao: There's only one class for channels at the moment. The other one is for user channels, i.e. DMs.

app/services/coms.js
raucao: I refactored the function to return early when there's no channel, or the message contains an error.

app/services/sockethub.js
silverbucket: OK this is a gap in the SH dynamic schema loading and validation, I've already been working on that area, and in the meantime I'm creating a SH issue with an exact reference to this piece of code to remove in the future.

raucao: OK, cool, thanks.

silverbucket (approved):

PR #309, opened 2026-06-02
Load sockethub client bundle from server at runtime

silverbucket:
## Summary

The sockethub server already serves its own pre-built client bundle at `/sockethub-client.js`. This bundle is built from the same codebase as the running server, so it always includes whatever object types the server supports.

Previously, hyperchannel installed `@sockethub/client` from npm and monkey-patched `validateActivity()` to bypass client-side schema validation. The bypass was necessary because the npm package's compiled schemas can lag behind the server when new object types are added (e.g. `room-info` in #308).

This PR instead loads the client bundle from the server at startup. The bundle sets `window.SockethubClient`, which is used to construct the client instance. This means the client-side schema validation always reflects the server's actual schema state — no bypass required.

The `@sockethub/client` npm package can be removed from `package.json` as a follow-up once anything else depending on it is audited.

## Test plan

- [x] Connect to a local sockethub server and confirm the app initialises correctly
- [x] Verify `window.SockethubClient` is set after the script loads
- [x] Confirm room-info queries work without any validation bypass

### DIFF
--- a/app/services/sockethub.js
+++ b/app/services/sockethub.js
@@ -7,26 +7,27 @@ export default class SockethubService extends Service {
   _platformContextMap = new Map();
 
   async initialize () {
-    await this._loadSockethubLibs();
-    const socket = window.io(config.sockethubURL, { path: '/sockethub' });
-    const client = new window.SockethubClient(socket);
-    await client.ready();
-    this.client = client;
-    this._buildPlatformContextMap();
+    return this.loadSockethubLibs(config.sockethubURL).then(async () => {
+      const socket = window.io(config.sockethubURL, { path: '/sockethub' });
+      const client = new window.SockethubClient(socket);
+      await client.ready();
+      this.client = client;
+      this._buildPlatformContextMap();
+    });
   }
 
-  async _loadSockethubLibs () {
-    await this._loadExternalScript(`${config.sockethubURL}/socket.io.js`);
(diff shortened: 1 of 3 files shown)


raucao:
This was changed to not load it from the Sockethub server in the last PR before today. What is the reasoning for 1) reverting the change already and 2) not reverting the commit that introduced the change, so the code is back to the original?

silverbucket:
> This was changed to not load it from the Sockethub server in the last PR before today. What is the reasoning for 1) reverting the change already and 2) not reverting the commit that introduced the change, so the code is back to the original?

Originally I wanted the type hinting from loading the client library, so I changed it to use the npm package. However, since then the client lib got dynamic client-side validation of platforms as they are defined in the running sockethub instance. 

I would have changed it in the previous PR but I confused the validation of extensions on incoming objects with the validation of object types in the platform schemas.

This PR not only goes back to loading the client lib from SH, but also keeps the npm package dependency and adds a globals.d.ts to link the typing from the package to the global sockethub window object. This way we get all the typing hints (at least sans any unpublished changes), while real-time validation works based on the sockethub code being run.

raucao:
OK, but then please use the same loading as in the original code, which assigned different properties to the script (module, async, e.g.).

silverbucket:
This works, we no longer need to override the client validation, and we do see a number of warnings from activity streams, these will be fixed in https://github.com/sockethub/sockethub/pull/1088

PR #313, opened 2026-06-14
Fix IRC attendance query resulting in name/presence flood

### LINKED ISSUE #312: IRC: queryAttendance omits target.name, causing a bare NAMES flood
## Summary

On IRC, `queryAttendance` sends a Sockethub `query`/`attendance` job whose `target` has only an `id` and no `name`. The Sockethub IRC platform builds the `NAMES` command from `target.name`, so the omission produces a **bare `NAMES`** (no channel argument). IRC servers answer a bare `NAMES` with the entire network channel list, flooding the connection with presence for every channel — the cause of sockethub/sockethub#1085 ("presence messages from rooms I had never joined").

Our client-side guard in `handlePresenceUpdate` (drops presence for unjoined channels) hides the symptom in the UI, but the flood still happens on the wire on every connect, since `queryAttendance` runs from `handleJoinCompleted` after each join.

## Root cause

`app/services/sockethub-irc.js` — `queryAttendance`:

```js
let msg = this.buildActivityObject(channel.account, {
  type: 'query',
  target: { id: channel.sockethubChannelId, type: 'room' }, // no `name`
  object: { type: 'attendance' }
});
```

Compare `join` (same file), which correctly sets `name: channel.name`. The Sockethub client normalizer does not derive `name` from `id`, so the platform receives `target.name === undefined` and emits `NAMES ` (bare).

## Fix

Set the channel name on the attendance target, mirroring `join`:

```js
target: { id: channel.sockethubChannelId, type: 'room', name: channel.name }
```

## Notes

- A complementary hardening fix is being made in Sockethub so a missing/invalid channel never produces a bare `NAMES` (rejects the query instead). With that in place this would no longer flood the network, but setting `name` here is still the correct client behavior.
- Refs: sockethub/sockethub#1085

silverbucket:
## Summary

On IRC, `queryAttendance` sent a Sockethub `query`/`attendance` job whose `target` had only an `id` and no `name`. The Sockethub IRC platform builds the `NAMES` command from `target.name`, so the omission produced a bare `NAMES` (no channel argument). IRC servers answer a bare `NAMES` with the entire network channel list, flooding the connection with presence for every channel on every join.

This sets `name` on the attendance target, mirroring what `join` already does.

Closes #312
Refs sockethub/sockethub#1085

### DIFF
--- a/app/services/sockethub-irc.js
+++ b/app/services/sockethub-irc.js
@@ -228,7 +228,7 @@ export default class SockethubIrcService extends Service {
   queryAttendance (channel) {
     let msg = this.buildActivityObject(channel.account, {
       type: 'query',
-      target: { id: channel.sockethubChannelId, type: 'room' },
+      target: { id: channel.sockethubChannelId, type: 'room', name: channel.name },
       object: {
         type: 'attendance'
       }


raucao (approved):
Works as advertised. 👍

PR #317, opened 2026-07-09
Update Sockethub ID for IRC channels, allow multiple `#` chars

raucao:
* Use `#channelname@host` for Sockethub channel IDs
* Allow multiple `#` characters as prefix, e.g. `##foobar@example.com`. Continue supporting URLs omitting the prefix (i.e. auto-prefix when no `#` are given).
* Also supports channel names with multiple `#` for the IRC logs now, but also has to be supported by loggers before it can work

refs https://github.com/sockethub/sockethub/pull/1131

### DIFF
--- a/app/components/channel-nav.hbs
+++ b/app/components/channel-nav.hbs
@@ -18,7 +18,7 @@
               {{/if}}
             {{else}}
               <LinkTo @route="channel" @model={{channel}} class="hc-sidebar-item">
-                #&thinsp;{{channel.shortName}}
+                #&thinsp;{{channel.sidebarName}}
               </LinkTo>
             {{/if}}
             <a {{on "click" (fn @onLeaveChannel channel)}} title="Leave {{channel.displayName}}"
--- a/app/models/base_channel.js
+++ b/app/models/base_channel.js
@@ -56,22 +56,30 @@ export default class BaseChannel {
         id = this.name;
         break;
       case 'IRC':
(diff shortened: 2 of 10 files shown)


silverbucket:
Here's what I came up with after a few review runs. I'm adding some tests which seem to confirm.

### P2 — Double decoding breaks channel navigation

In `app/routes/base_channel.js:17`, Ember has already decoded `params.slug`. Calling `decodeURIComponent()` again throws `URI malformed` for `#100%`, and changes `#foo%23bar` into `#foo#bar`, potentially joining the wrong channel.

Previously, already-loaded channels were matched before decoding. Use the router-provided value directly.

### P2 — Domain parsing breaks IRC channels containing `/`

In `app/utils/domain-from-id.js:2–4`, stripping everything after the first `/` makes `#foo/bar@irc.libera.chat` return no domain. Sockethub accepts this channel ID.

This breaks sidebar grouping and redirects deep links for unloaded channels to the first channel. Resource stripping needs to distinguish XMPP resource suffixes from slashes in IRC channel names.

**Validation:** Four regression tests using Ember’s actual router in Chrome pass against the PR base and fail against the PR head, covering both issues.

silverbucket (changes requested):
Added some feedback and failing tests.

raucao:
@silverbucket Thanks! Fixed the remaining issues and refactored and moved the tests into the correct suites.

silverbucket (approved):
LGTM!

